PR #49, opened 2026-05-18
Add deepseek_v4_flash.py SglangRecipe

atoniolo76:
Add a [DeepSeek V4 Flash](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash&ved=2ahUKEwj8y_C1ssOUAxW-L1kFHTjWL2QQFnoECCAQAQ&usg=AOvVaw2yU3TkZwF5VqqgipGiFWW-) SGLang recipe.

Default config uses 4:B200 devices since DeepSeek V4 uses FP4 quantization for the MoE expert parameters. 

## Checklist

## Outside contributors


joyliu-q (commented):
Can we add a tutorial either distilling from this or using it somewhere?

Want to make sure this example works before merging, and we currently have a loop that daily tests all existing tutorials.

PR #62, opened 2026-05-21
Bump SGLang Image Version

atoniolo76:
Fixes error with image build using old slang image `v0.5.12-cu130`.

## Checklist

## Outside contributors


joyliu-q (commented):
can u pin it to an actual version that works and is not nightly

joyliu-q (approved):

PR #66, opened 2026-05-26
Add model-aware response parsing via parse_response() → ParsedResponse

joyliu-q:
Adds structured model output parsing to `ModelConfig`, following the same pattern as vLLM's `ToolParser.extract_tool_calls()` and HuggingFace's `tokenizer.parse_response()` — one method that takes raw text and returns structured output.

### API

```python
@dataclass
class ToolCall:
    name: str
    arguments: dict[str, Any]

@dataclass
class ParsedResponse:
    content: str                    # clean text, no thinking/template/tool blocks
    tool_calls: list[ToolCall]
    thinking: str = ""              # extracted reasoning, if any

class ModelConfig:
    response_parser: ResponseParser | None = None

    def parse_response(self, text: str) -> ParsedResponse:
        """Delegates to response_parser when set, otherwise returns text as-is."""
```

### Qwen3 implementation

`parse_qwen3_response(text) → ParsedResponse` is a standalone function that handles:
- `<think>…</think>` reasoning blocks → `ParsedResponse.thinking`
- `<|im_start|>`/`<|im_end|>` chat-template delimiters → stripped
- `<tool_call>{"name": …, "arguments": …}</tool_call>` → `ParsedResponse.tool_calls`

All 7 Qwen3 model classes set `response_parser = parse_qwen3_response` as a class attribute.

### ParsedResponse in the Eval UI

- `EvalRowResult` now has an optional `parsed_response` field (dict with `content`, `thinking`, `tool_calls`)
- The eval dashboard renders parsed fields separately when available:
  - **Thinking** — dimmed with orange left border
  - **Content** — clean response text
  - **Tool calls** — structured JSON
  - Falls back to raw response display for rows without `parsed_response`

### Exports

`ToolCall`, `ParsedResponse`, and `parse_qwen3_response` are exported from the top-level package and added to the API reference manifest.

## Checklist

Link to Devin session: https://modal.devinenterprise.com/sessions/0b72965b6f7d45b78fbf972b67ff7263
Requested by: @joyliu-q


modal_training_gym/common/eval.py
```diff
@@ -57,6 +57,7 @@ class EvalRowResult(BaseModel):
     score: float
     response: str = ""  # TODO, this doesn't have to be a string
     prompt: str = ""
+    parsed_response: dict[str, Any] | None = None
```
joyliu-q: is the type dict[str, Any]?

modal_training_gym/common/models/base.py
```diff
@@ -114,17 +116,42 @@ def to_megatron_args(self) -> list[str]:
...
+    arguments: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class ParsedResponse:
+    """Structured result of parsing raw model output."""
+
+    content: str
+    tool_calls: list[ToolCall] = field(default_factory=list)
+    thinking: str = ""
```
joyliu-q: this should be optional

joyliu-q: no i mean that

thinking: str | None = None

modal_training_gym/common/models/base.py
```diff
@@ -114,17 +116,42 @@ def to_megatron_args(self) -> list[str]:
...
+    name: str
+    arguments: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class ParsedResponse:
+    """Structured result of parsing raw model output."""
+
+    content: str = ""
+    tool_calls: list[ToolCall] | None = None
```
joyliu-q: tool_calls: list[ToolCall]  = []

modal_training_gym/common/models/base.py
```diff
@@ -114,17 +116,42 @@ def to_megatron_args(self) -> list[str]:
...
+    thinking: str | None = None
+
+
+ResponseParser = Callable[[str], ParsedResponse]
```
joyliu-q, line 136: What happens if the response is streaming

PR #80, opened 2026-05-30
Support slime sglang custom params

atoniolo76:
Adds a custom sglang client for passing arbitrary parameters like `top_logprobs_num` into request payload. Provides replacements for Slime reward functions, which make a hard-coded request to a sglang engine, to support arbitrary parameters in non-custom reward functions. Adds a new request interface for passing parameters into payload for user's custom reward functions.

## Checklist

## Outside contributors


modal_training_gym/deploy_recipes/sglang_recipe/serve_sglang.py
joyliu-q: Why min_containers 1 here? This accumulates costs for our users once deployed, and requires the app to be explicitly deleted as a teardown process

modal_training_gym/common/sglang_client.py
```diff
@@ -0,0 +1,86 @@
+"""Generic async client for SGLang ``/generate`` endpoints.
```
joyliu-q, line 1: Can we move this to be outside of slime so it can be reused?

atoniolo76: Moved to `common`

modal_training_gym/deploy_recipes/sglang_recipe/serve_sglang.py
atoniolo76: Fine to remove, a scale down window of 10 minutes should be adequate for most workloads.

joyliu-q (approved):
lgtm ty king

PR #84, opened 2026-06-01
Fix submit_and_tail killing jobs when Ray log stream disconnects early.

atoniolo76:
Ray's tail_job_logs() can silently drop the gRPC/WebSocket stream while the job is still RUNNING (ray-project/ray#57002, #60665). The old code immediately checked job status after the stream ended and raised if it wasn't SUCCEEDED — killing a healthy training run.

Now submit_and_tail polls until the job reaches a terminal state (SUCCEEDED/FAILED/STOPPED) and reconnects the log stream on each iteration, so transient disconnects no longer abort training.

## Checklist

## Outside contributors

### DIFF
--- a/modal_training_gym/common/ray_cluster.py
+++ b/modal_training_gym/common/ray_cluster.py
@@ -250,6 +250,7 @@ class ModalRayCluster:
         entrypoint: str,
         *,
         runtime_env: dict | None = None,
+        max_retries: int = 10,
     ) -> str:
         """Submit a Ray job, stream its logs to stdout, and return the final status."""
         if not self.is_head:
@@ -262,6 +263,7 @@ class ModalRayCluster:
         print(f"Submitted Ray job: {job_id}")
 
         _TERMINAL = {"SUCCEEDED", "FAILED", "STOPPED"}
+        retry_count = 0
 
         while True:
             log_stream = self._client.tail_job_logs(job_id)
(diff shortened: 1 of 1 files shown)


modal_training_gym/common/ray_cluster.py
```diff
@@ -260,18 +260,31 @@ async def submit_and_tail(
...
+                    print(line, end="", flush=True)
+
+            status = self._client.get_job_status(job_id).value
+            if status in _TERMINAL:
+                break
+            print(
+                f"\n[ray] Log stream ended but job {job_id} is still {status}; "
+                "reconnecting in 5s..."
+            )
+            await asyncio.sleep(5)
```
joyliu-q: Let's do it with max retries

modal_training_gym/common/ray_cluster.py
```diff
@@ -250,6 +250,7 @@ async def submit_and_tail(
         entrypoint: str,
         *,
         runtime_env: dict | None = None,
+        max_retries: int = 10,
```
joyliu-q: Let's do 35or something more chill, or max retries 10 but sleep for 2s intervals

modal_training_gym/common/ray_cluster.py
```diff
@@ -260,18 +261,38 @@ async def submit_and_tail(
...
-                print(line, end="", flush=True)
-        else:
-            for line in log_stream:
-                print(line, end="", flush=True)
-        status = self._client.get_job_status(job_id).value
+
+        _TERMINAL = {"SUCCEEDED", "FAILED", "STOPPED"}
+        retry_count = 0
+
+        while True:
```
joyliu-q: Nit: can be a `while retry_count ...` instead of `while True`

joyliu-q (approved):

PR #90, opened 2026-06-03
New ID generation for training gym

cskitty:
Added new simple id generation for run IDs for training, deployment, and eval.

Run ID:
Format: {word}-{word}-{hash}
Hash is sha256(config fingerprint + ":" + attempt_index)[:5]

Eval ID:
{descriptive-slug}-{6-char}

Screenshot 2026-06-03 at 3 05 33 PM

### DIFF
--- a/modal_training_gym/common/deployment.py
+++ b/modal_training_gym/common/deployment.py
@@ -23,7 +23,11 @@ from modal_training_gym.common.checkpoint import (
     CheckpointType,
     convert_checkpoint_to_hf,
 )
-from modal_training_gym.common.ids import GymObjectId, config_fingerprint, stable_readable_id
+from modal_training_gym.common.ids import (
+    GymObjectId,
+    config_fingerprint,
+    stable_readable_id,
+)
 from modal_training_gym.common.models import ModelConfig
 from modal_training_gym.common.modal_urls import modal_app_dashboard_url
 from modal_training_gym.deploy_recipes.sglang_recipe import SglangRecipe
@@ -128,7 +132,9 @@ class DeploymentConfig:
         if not self.served_model_name:
             self.served_model_name = default_slug
 
-        deployment_stable_id = self._new_deployment_id(recipe=recipe, model_path=model_path)
+        deployment_stable_id = self._new_deployment_id(
+            recipe=recipe, model_path=model_path
+        )
         deployment_id = deployment_stable_id.value
         checkpoints_volume = self._checkpoints_volume_name()
         checkpoints_mount_path = self._checkpoints_mount_path()
--- a/modal_training_gym/common/deployment.py
+++ b/modal_training_gym/common/deployment.py
@@ -15,7 +15,6 @@ from dataclasses import dataclass
 from enum import Enum
 import inspect
 import threading
-import uuid
 
 from pydantic import BaseModel, ConfigDict, field_serializer, field_validator
 
@@ -24,6 +23,7 @@ from modal_training_gym.common.checkpoint import (
     CheckpointType,
     convert_checkpoint_to_hf,
 )
+from modal_training_gym.common.ids import config_fingerprint, unique_readable_id
 from modal_training_gym.common.models import ModelConfig
 from modal_training_gym.common.modal_urls import modal_app_dashboard_url
 from modal_training_gym.deploy_recipes.sglang_recipe import SglangRecipe
@@ -79,18 +79,27 @@ class DeploymentConfig:
             )
         return model_path
 
-    def _new_deployment_id(self) -> str:
-        model_name = (
-            self.served_model_name or self.model.model_name or self.model.model_path
+    def _new_deployment_id(
+        self,
+        *,
+        recipe: VllmRecipe | SglangRecipe,
+        model_path: str,
+    ) -> tuple[str, str, int]:
+        fingerprint = config_fingerprint(
+            "deployment",
+            self.model,
+            self.checkpoint,
+            recipe,
+            self.app_name,
+            self.served_model_name,
+            model_path,
         )
-        if self.checkpoint is not None:
-            model_name = f"{model_name}-{self.checkpoint.name}"
-
-        recipe_name = (
-            self.recipe.recipe_type.value if self.recipe is not None else "sglang"
+        deployment_id, attempt_index = unique_readable_id(
+            MetadataStore.DEPLOYMENTS,
+            fingerprint,
+            id_key="deployment_id",
         )
-
-        return f"{model_name}.{recipe_name}.{uuid.uuid4().hex[:4]}"
+        return deployment_id, fingerprint, attempt_index
 
     # TODO: add _merge_recipe for deployment configs
     def serve(self) -> "ModelDeployment":
@@ -120,7 +129,11 @@ class DeploymentConfig:
         if not self.served_model_name:
             self.served_model_name = default_slug
 
-        deployment_id = self._new_deployment_id()
+        (
+            deployment_id,
+            deployment_fingerprint,
+            attempt_index,
+        ) = self._new_deployment_id(recipe=recipe, model_path=model_path)
         checkpoints_volume = self._checkpoints_volume_name()
         checkpoints_mount_path = self._checkpoints_mount_path()
 
@@ -191,6 +204,8 @@ class DeploymentConfig:
             modal_app_url=modal_app_url or modal_app_dashboard_url(modal_app_id),
             url=url,
             status=DeploymentStatus.RUNNING.value,
+            config_fingerprint=deployment_fingerprint,
+            attempt_index=attempt_index,
         )
         deployment.save()
         return deployment
@@ -240,6 +255,8 @@ class ModelDeployment(BaseModel):
     modal_app_url: str = ""
     url: str
     status: str = DeploymentStatus.RUNNING.value
+    config_fingerprint: str = ""
+    attempt_index: int = 0
 
     @field_validator("deployment_config", mode="before")
     @classmethod
--- a/modal_training_gym/common/deployment.py
+++ b/modal_training_gym/common/deployment.py
@@ -23,11 +23,7 @@ from modal_training_gym.common.checkpoint import (
     CheckpointType,
     convert_checkpoint_to_hf,
 )
-from modal_training_gym.common.ids import (
-    GymObjectId,
-    config_fingerprint,
-    stable_readable_id,
-)
+from modal_training_gym.common.ids import create_hash
 from modal_training_gym.common.models import ModelConfig
 from modal_training_gym.common.modal_urls import modal_app_dashboard_url
 from modal_training_gym.deploy_recipes.sglang_recipe import SglangRecipe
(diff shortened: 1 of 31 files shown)


modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+                if not str(k).startswith("_") and not callable(v)
+            },
+        }
+    return repr(value)
+
+
+def config_fingerprint(*parts: Any) -> str:
+    try:
+        payload = json.dumps(canonicalize(parts), sort_keys=True, separators=(",", ":"))
+    except Exception:
```
joyliu-q: Instead of a generic `Exception`, can we use something more specific? Also, why would this ever throw an error?

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+    return re.sub(r"-{2,}", "-", slug)
+
+
+def canonicalize(value: Any) -> Any:
+    if value is None or isinstance(value, str | int | float | bool):
+        return value
+    if isinstance(value, Enum):
+        return value.value
+    if isinstance(value, Path):
+        return str(value)
```
joyliu-q: Is it the right idea to return the stringified value for a Path? Why would you ever pass a Path in

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+
+_NON_ALNUM_RE = re.compile(r"[^a-z0-9]+")
+
+
+def slugify(value: str) -> str:
+    slug = _NON_ALNUM_RE.sub("-", value.lower()).strip("-")
+    return re.sub(r"-{2,}", "-", slug)
+
+
+def canonicalize(value: Any) -> Any:
```
joyliu-q: Generally discourage against type Any. This shows that this function is doing too much / too broad.

Also the name of the function does not explain to me what it is actually doing, where it will be used, why would I use it, etc

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+from pydantic import BaseModel
+
+from modal_training_gym.utils.metadata import MetadataStore, vol_list
+
+_NON_ALNUM_RE = re.compile(r"[^a-z0-9]+")
+
+
+def slugify(value: str) -> str:
+    slug = _NON_ALNUM_RE.sub("-", value.lower()).strip("-")
+    return re.sub(r"-{2,}", "-", slug)
```
joyliu-q: This seems like it can be an inline instead of a fn

modal_training_gym/common/deployment.py
```diff
@@ -79,18 +79,27 @@ def _resolved_model_path(self) -> str:
...
 
-    def _new_deployment_id(self) -> str:
-        model_name = (
-            self.served_model_name or self.model.model_name or self.model.model_path
+    def _new_deployment_id(
+        self,
+        *,
+        recipe: VllmRecipe | SglangRecipe,
+        model_path: str,
+    ) -> tuple[str, str, int]:
```
joyliu-q: Can we keep the return type to be a str, or make a new object `GymObjectId` that has these fields named?

Generally returning tuples are a bad pattern because when you unpack them, they don't come with explanations on what each of the results are.

modal_training_gym/common/deployment.py
```diff
@@ -240,6 +255,8 @@ class ModelDeployment(BaseModel):
     modal_app_url: str = ""
     url: str
     status: str = DeploymentStatus.RUNNING.value
+    config_fingerprint: str = ""
+    attempt_index: int = 0
```
joyliu-q: Is `attempt_index` an index, or more like a count? What happens when we have a race condition (is this value monotonically increasing and how do we ensure that)

modal_training_gym/common/eval.py
```diff
@@ -195,14 +202,21 @@ def eval_fn(
...
             eval_fn_name = _callable_name(self.eval_fn or self.eval_response_fn)
-            self.eval_config_id = (
-                f"{class_name}.{dataset_name}.{eval_fn_name}.{uuid4().hex[:4]}"
+            fingerprint = config_fingerprint(
+                "eval-config",
+                class_name,
+                self.dataset,
+                eval_fn_name,
+                self.prompt_column,
+                self.generate_kwargs,
```
joyliu-q: Why are these the fields we are passing in? Generally `config_fingerprint` seems to be typed too loosely and the fields we use to derive a fingerprint seems a bit adhoc

modal_training_gym/common/ids.py
joyliu-q: My guess is that we will not need half of these and/or should not support them. This is part of the issue with config_fingerprint being typed too loosely as mentioned above

modal_training_gym/common/ids.py
joyliu-q: Why are we dumping this into a json just to hash it

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+        payload = repr(parts)
+    return hashlib.sha256(payload.encode()).hexdigest()[:12]
+
+
+def id_hash(fingerprint: str, attempt_index: int, *, length: int = 5) -> str:
+    payload = f"{fingerprint}:{attempt_index}"
+    return hashlib.sha256(payload.encode()).hexdigest()[:length]
+
+
+def random_word_slug() -> str:
```
joyliu-q: Why import locally instead of at top level? Instead of taking this current comment feedback and just moving the import, please look into what the difference of the two is, write explanation below, and then make any necessary changes or defend this decision.

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+
+def config_fingerprint(*parts: Any) -> str:
+    try:
+        payload = json.dumps(canonicalize(parts), sort_keys=True, separators=(",", ":"))
+    except Exception:
+        payload = repr(parts)
+    return hashlib.sha256(payload.encode()).hexdigest()[:12]
+
+
+def id_hash(fingerprint: str, attempt_index: int, *, length: int = 5) -> str:
```
joyliu-q: Ditto as `slugify`, this does not need to be a separate function, which hides what it is actually doing, and should be done inline. 

It's also only called once.

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+def random_word_slug() -> str:
+    import randomname
+
+    slug = slugify(randomname.get_name(sep="-"))
+    if slug.count("-") >= 1:
+        return slug
+    return f"run-{secrets.token_hex(3)}"
+
+
+def _metadata_records(store: MetadataStore) -> list[dict[str, Any]]:
```
joyliu-q: Any

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+    try:
+        return vol_list(store)
+    except Exception:
+        return []
```
joyliu-q: Why would it ever return an exception, and should we log in that case?

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _metadata_value(record: dict[str, Any], key: str) -> Any:
```
joyliu-q: Any

modal_training_gym/common/ids.py
```diff
@@ -0,0 +1,135 @@
...
+def next_attempt_index(store: MetadataStore, fingerprint: str) -> int:
+    attempts: list[int] = []
+    for record in _metadata_records(store):
+        if _metadata_value(record, "config_fingerprint") != fingerprint:
+            continue
+        attempt = _metadata_value(record, "attempt_index")
+        if isinstance(attempt, int):
+            attempts.append(attempt)
+        elif isinstance(attempt, str) and attempt.isdigit():
+            attempts.append(int(attempt))
```
joyliu-q: Why would there be multiple types that attempts could be? Shouldn't this be uniform from the start

modal_training_gym/common/train.py
```diff
@@ -37,12 +38,29 @@ class TrainConfig:
...
+    _training_run_id: str = _dc.field(default="", init=False, repr=False)
+    _config_fingerprint: str = _dc.field(default="", init=False, repr=False)
+    _attempt_index: int = _dc.field(default=0, init=False, repr=False)
 
     # ── Public API ────────────────────────────────────────────────────────────
 
     @property
     def training_run_id(self) -> str:
-        return f"{self.model.model_name}.{self.dataset.name}.{uuid.uuid4().hex[:4]}"
+        # Computed once and cached: a stable id for the lifetime of this config.
```
joyliu-q: Explain why we want a stable id here in a comment

joyliu-q (commented):
Can we add some tests on id generation? No worries if not

modal_training_gym/common/ids.py
cskitty: I believe the local import was there so randomname failures only show when ID generation runs and other code could import this module without needing randomname installed. I think this actually isn't that useful here since this file should generally only be called when the ID generation is happening anyway so will remove.

modal_training_gym/common/ids.py
cskitty: Fixed exception, I thought that would be an okay way to standardize the hash format

modal_training_gym/common/deployment.py
```diff
@@ -240,6 +257,8 @@ class ModelDeployment(BaseModel):
     modal_app_url: str = ""
     url: str
     status: str = DeploymentStatus.RUNNING.value
+    config_fingerprint: str = ""
+    id_created_at: int = 0
```
joyliu-q: Why is this an int, not a datetime?

modal_training_gym/common/eval.py
```diff
@@ -109,6 +110,8 @@ class EvalResult(BaseModel):
     eval_id: str
     eval_config_id: str
     deployment_id: str
+    config_fingerprint: str = ""
+    id_created_at: int = 0
```
joyliu-q: ditto

modal_training_gym/common/deployment.py
```diff
@@ -240,6 +257,8 @@ class ModelDeployment(BaseModel):
     modal_app_url: str = ""
     url: str
     status: str = DeploymentStatus.RUNNING.value
+    config_fingerprint: str = ""
```
joyliu-q: When will this default to `""`?

modal_training_gym/common/eval.py
```diff
@@ -195,14 +198,29 @@ def eval_fn(
...
+                    "-",
+                    re.sub(
+                        r"[^a-z0-9]+",
+                        "-",
+                        f"{class_name}-{dataset_name}-{eval_fn_name}".lower(),
+                    ).strip("-"),
+                )
+                or "eval-config"
             )
+            slug = raw_slug
```
joyliu-q: This can be one line

joyliu-q (approved):

PR #103, opened 2026-06-05
Add MultimodalDataset and Audio RL Example

BenCowen:
# Qwen3-ASR audio RL example + supporting gym features

Adds an end-to-end audio GRPO example — Qwen3-ASR-1.7B (trained with reward = −WER), plus the reusable gym pieces it needed.

## Gym features

- **`MultimodalDataset`**: encapsulates the multimodal-key and media-column wiring for image/audio/video RL (emits `--multimodal-keys`).
- **`disable_clustered`** (`_maybe_clustered`): runs single-node without the `@clustered` decorator, for better single-node scheduling.
- **`disable_hf_conversion`**: skips the post-training HF export for models whose bridge export isn't supported.
- **`_ship_callable` fix**: sets the slime arg path when a `custom_rm_function` or `custom_generate_function` is defined outside the entrypoint module. Before this the file was shipped but the path was left unset, so it silently fell back to a rule-based RM.
- **Audio evals panel**: adds an `<audio>` player plus reference text and WER to eval rows in the dashboard. Keyed off `row.metadata`, so it's inert for non-audio evals and needs no backend change.

### Evals audio panel
[Example deployed here](https://modal-labs-ben-dev--audio-evals-dashboard-fastapi-app.modal.run/evals), and screenshot attached.
image

## Example

`tutorials/rl/005_audio_asr/` — LibriSpeech audio → SGLang `/v1/audio/transcriptions` rollout → −WER reward → GRPO. Default run is a full 8×H100 node; `--minimal` runs a 2×H100 demo. The trained model is exported to a standard HF checkpoint.

A big part of the contribution is `_native_qwen3asr_compat.py`, which applies idempotent image-build workarounds for four upstream gaps on the gym's current stack: the bridge config validate-order, slime processor loading, the bridge `pg_collection` default, and the MB→HF export converter (audio tower). 

## Tests

- `tests/test_multimodal_passthrough.py` — unit test for `MultimodalDataset` passthrough.

## Follow-ups

- Multinode: rollout discovery is in place; cross-node save and GRPO stability on larger/noisier data are follow-ups.
- The HF export is confirmed structurally complete but not yet load-and-transcribe verified.

# Checklist


modal_training_gym/common/dataset.py
```diff
@@ -544,3 +544,70 @@ def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
...
+    or base64 data — whatever the serving engine accepts; the gym never
+    inspects them.
+
+    Pass ``rows=[{"prompt": str, "media": list, "label": Any}, ...]`` or
+    subclass and override ``rows()``.
+    """
+
+    input_key: str = "prompt"
+    label_key: str = "label"
+    modality: str = "image"  # "image" | "audio" | "video"
```
joyliu-q: Instead of defaulting to image, let's do audio here since thats our working example?

joyliu-q: This can also be a python enum or typed not as a str but must be "image" | "audio" | "video", which will make it a type-check time validation

modal_training_gym/common/dataset.py
```diff
@@ -544,3 +544,70 @@ def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
...
+            raise ValueError("media_column must differ from input_key and label_key")
+        # The whole feature in one line: name the media column for the framework.
+        self.multimodal_keys = {self.modality: self.media_column}
+        self._rows = list(rows or [])
+        if not self.dataset_id:
+            self.dataset_id = f"mm-{self.modality}-{uuid.uuid4()}"
+        self._validate()
+
+    def rows(self) -> list[dict[str, Any]]:
+        return self._rows
```
joyliu-q: nit this is a `@property`

modal_training_gym/common/dataset.py
```diff
@@ -544,3 +544,70 @@ def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
...
+
+    def _to_row(self, r: dict[str, Any]) -> dict[str, Any]:
+        media = r["media"]
+        return {
+            self.input_key: r["prompt"],
+            self.media_column: list(media) if isinstance(media, (list, tuple)) else [media],
+            self.label_key: r["label"],
+        }
```
joyliu-q: Can we type this better than `Any`

modal_training_gym/common/dataset.py
```diff
@@ -544,3 +544,70 @@ def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
...
+class MultimodalDataset(DatasetConfig):
+    """Modality-agnostic dataset for image / audio / video RL.
+
+    Each row pairs a text ``prompt`` with one or more ``media`` items and a
+    ``label``. ``prepare()`` writes the media verbatim into a column named by
+    ``media_column`` (default ``"<modality>s"``), and the column is surfaced to
+    the trainer/rollout via ``multimodal_keys`` (``{modality: media_column}``,
+    e.g. slime's ``--multimodal-keys``). Media items may be URLs, local paths,
+    or base64 data — whatever the serving engine accepts; the gym never
+    inspects them.
```
joyliu-q, line 558: If they can take any format, how does our platform visualize them for debugging?

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -71,18 +71,34 @@
...
-)
+_PATCH_VALIDATION_B64 = encode_patch("patch_validation")
+_PATCH_MEGATRON_BRIDGE_B64 = encode_patch("patch_megatron_bridge")
+_PATCH_TORCH_LOAD_B64 = encode_patch("patch_torch_load")
+_PATCH_GLOBAL_PLAN_B64 = encode_patch("patch_global_plan")
+_PATCH_CHECKPOINT_SAVE_B64 = encode_patch("patch_checkpoint_save")
+_PATCH_ADVANTAGES_B64 = encode_patch("patch_advantages")
+
+
+def _maybe_clustered(use_clustered: bool, size: int, *, rdma: bool):
```
joyliu-q: Return type

joyliu-q: Open question: is there a better name than `_maybe_clustered`

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -273,6 +271,10 @@ def _ship_callable(
                 remote_path=f"/root/{fn_module_name}.py",
                 copy=True,
             )
+            # Point the slime arg at the shipped module's symbol. Without this the
+            # file is shipped but the path stays unset, so a custom_rm_function
+            # defined outside the entrypoint silently falls back to rule-based RM.
+            set_path(f"{fn_module_name}.{getattr(fn, '__name__', fallback_name)}")
```
joyliu-q, line 367: Great catch!

modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -98,6 +100,11 @@ class SlimeRecipe(BaseTrainRecipe):
     memory: int | tuple[int, int] | None = None
     cloud: str | None = None
     region: str | None = None
+    disable_clustered: bool = False
```
joyliu-q: Why would this value be set

modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -98,6 +100,11 @@ class SlimeRecipe(BaseTrainRecipe):
...
+    # Skip the post-training megatron→HF checkpoint conversion. Useful when the model
+    # trains via the bridge but its MB→HF export is unsupported (e.g. Qwen3-ASR's
+    # audio tower), so a failed optional export shouldn't fail a successful run.
+    disable_hf_conversion: bool = False
```
joyliu-q: Can this log if set for debugging

joyliu-q: Context is that it seems easy for an agent to set this value not realize and shoot itself in the foot

tests/test_multimodal_passthrough.py
```diff
@@ -0,0 +1,62 @@
+"""The multimodal passthrough: a dataset names its media column, the recipe
```
joyliu-q, line 1: Thanks for adding a test!

tutorials/rl/005_audio_asr/_native_qwen3asr_compat.py
```diff
@@ -0,0 +1,177 @@
...
+     slime has initialized model-parallel state — so default to it instead of
+     raising. (Other megatron-bridge models accept a None pg_collection; Qwen3-ASR
+     is the outlier.)
+
+Run at image build:  python _native_qwen3asr_compat.py
+"""
+import pathlib
+
+
+def _patch_bridge_config() -> None:
```
joyliu-q: This can be a patch file we add globally in our system (see how we patch existing slime), but we conditionally patch this if its an audio model.

modal_training_gym/train_recipes/slime_recipe/recipe.py
BenCowen: basically, `@clustered` has a restricted pool. So I preferred to turn it off when using <=1 node for faster dev cycle. It also "saves" those nodes for functions that actually need clusters.

BenCowen: maybe it should be set automagically based on how many nodes are needed

modal_training_gym/train_recipes/slime_recipe/recipe.py
BenCowen: i believe this is actually vestigial and no longer needed

modal_training_gym/common/dataset.py
BenCowen: TODO added here:
https://github.com/modal-projects/training-gym/pull/103/changes#diff-19d2240f232488d3441ecf8713792950fb2a29fbdba8cd432c9d38a756d7f5a9R566

modal_training_gym/common/models/__init__.py
```diff
@@ -38,4 +39,5 @@
     "ToolCall",
     "parse_qwen3_response",
     "Qwen3_6_35B",
+    "Qwen3ASR",
```
joyliu-q: Can you specify this by the name: Which Qwen3 ASR?

Also name should be
`Qwen3_ASR_1_7B`  following other model naming conventions

modal_training_gym/common/models/base.py
```diff
@@ -131,6 +131,11 @@ class ModelArchitecture:
     moe_aux_loss_coeff: float | None = None
     megatron_spec: list[str] | None = None
     megatron_model_type: str = ""
+    # Names of compat patch files (in frameworks/slime/modal_helpers/patches/) to
+    # apply at image build for this model — for upstream gaps a model needs until
+    # they're fixed upstream (e.g. Qwen3-ASR's bridge/processor/export shims). The
+    # launcher applies these only when this model is used.
+    compat_patches: list[str] | None = None
```
joyliu-q: This only applies to this model when it's used on slime though, right? If we used miles or tinker, these patches would be obsolete and not mean anything.

In this case, it doesn't make sense to attach this metadata to the *model definition*, which should only include model specs.

So a better implementation would be to just remove this field entirely from here and add these patches automatically in the SlimeRecipe for Qwen3_ASR_...

modal_training_gym/common/models/qwen3_asr.py
```diff
@@ -0,0 +1,61 @@
...
+loading, bridge pg_collection, MB→HF export of the audio tower). Each should be
+reported upstream; once fixed there, the corresponding patch can be dropped.
+"""
+
+from __future__ import annotations
+
+from .base import HFModelConfiguration, ModelArchitecture
+
+
+class Qwen3ASR(HFModelConfiguration):
```
joyliu-q: Rename, as mentioned above

modal_training_gym/frameworks/slime/modal_helpers/patches/model_specific_patches/qwen3_asr/patch_qwen3_asr_bridge_config.py
```diff
@@ -0,0 +1,39 @@
+"""Compat shim: guard megatron-bridge's Qwen3-ASR config get_text_config.
```
joyliu-q, line 1: All the asr patches should be in a folder
`/slime/modal_helpers/patches/model_specific_patches/qwen3_asr/...`

joyliu-q: Also name should be
`patch_qwen3_asr_bridge_config` not `qwen3asr`

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -85,6 +85,30 @@
 _PATCH_STOP_TOKEN_DIAG_B64 = encode_patch("patch_stop_token_diagnostic", _SLIME_PATCHES)
 
 
+def _clustered_if(
```
joyliu-q: This seems like it applies to both slime and miles. Can we move this to the `common` folder and import from both

modal_training_gym/rollouts/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Gym-provided slime rollouts."""
+
+from .audio_transcription import transcription_rollout
+
+__all__ = ["transcription_rollout"]
```
joyliu-q: Why are we adding a rollout folder? Shouldn't `audio_transcription.py` live in `common`?

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
+"""Custom slime rollout that drives an ASR model via /v1/audio/transcriptions.
```
joyliu-q: Is this slime specific? If it is, it must go into `slime` folder
But this seems like a `common` functionality.

All of the methods use slime's `Sample` as an input, but we can factor out the common code so the slime audio function just does:

'Sample' --> "data" --> (_extract_audio_bytes) --> bytes

_extract_audio_bytes does "data" --> (_extract_audio_bytes) --> bytes

and in slime we add fn that does 'Sample' --> "data"

joyliu-q: Also this file seems like ai slop

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
...
+    ip = getattr(args, "sglang_router_ip", None)
+    port = getattr(args, "sglang_router_port", None)
+    if not ip:
+        return None
+    ip = str(ip)
+    host = ip if ip.startswith("http") else f"http://{ip}"
+    return f"{host}:{port}" if port else host
+
+
+def _candidate_engine_bases(args: Any) -> list[str]:
```
joyliu-q: Feels like slop. Also args is type any which makes me feel like it's too generic

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
...
+                continue
+            bases = [u if u.startswith("http") else f"http://{u}" for u in urls]
+            if bases:
+                _ENGINE_BASES = bases
+                return _ENGINE_BASES
+    # single node / router unavailable: probe the local node's engines.
+    return _candidate_engine_bases(args)
+
+
+_TOKENIZER = None
```
joyliu-q: Global tokenizer variable? Why?

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
...
+    """Cached tokenizer (with Qwen3-ASR chat template injected by our slime patch)."""
+    global _TOKENIZER
+    if _TOKENIZER is None:
+        hf = getattr(args, "hf_checkpoint", None) or getattr(args, "model_name", None)
+        try:
+            from slime.utils.processing_utils import load_tokenizer
+
+            _TOKENIZER = load_tokenizer(hf, trust_remote_code=True)
+        except Exception:
+            from transformers import AutoTokenizer
```
joyliu-q: Try except seems brittle

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
...
+    """Cached Qwen3-ASR processor (WhisperFeatureExtractor + tokenizer) — produces
+    mel input_features AND the prompt with <audio_pad> expanded to the audio
+    encoder's output length, so audio embeds align with token positions."""
+    global _PROCESSOR
+    if _PROCESSOR is None:
+        hf = getattr(args, "hf_checkpoint", None) or getattr(args, "model_name", None)
+        from sglang.srt.configs.qwen3_asr import Qwen3ASRProcessor
+
+        _PROCESSOR = Qwen3ASRProcessor.from_pretrained(hf)
+    return _PROCESSOR
```
joyliu-q: Can we not globally cache a processor in our training library

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+# Qwen3-ASR audio placeholder: the processor expands the single <|audio_pad|> to N
+# tokens (N = audio-encoder output length for the clip). Must appear in the prompt
+# TEXT — and the raw audio data-URI must NOT, or it tokenizes into ~300K-900K text
+# tokens (base64 length scales with clip duration) and the actor OOMs.
+_AUDIO_PLACEHOLDER = "<|audio_start|><|audio_pad|><|audio_end|>"
```
joyliu-q: This is Qwen3-ASR model specific right? Should this logic be moved into model specific code

modal_training_gym/rollouts/audio_transcription.py
```diff
@@ -0,0 +1,425 @@
...
+    for m in msgs:
+        if m["role"] == "user":
+            m["content"] = (_AUDIO_PLACEHOLDER + "\n" + m["content"]).strip()
+            break
+    else:
+        msgs.insert(0, {"role": "user", "content": _AUDIO_PLACEHOLDER})
+
+    try:
+        return tok.apply_chat_template(msgs, tokenize=False, add_generation_prompt=True)
+    except Exception:
```
joyliu-q: Exception is too general here, is there a specific Exception we are expecting, and why would it throw? In those cases, why would the user want their training run to fall back here instead of failing

modal_training_gym/train_recipes/slime_recipe/qwen3_asr.py
```diff
@@ -0,0 +1,77 @@
...
+
+from pydantic import ConfigDict
+from pydantic.dataclasses import dataclass
+
+from modal_training_gym.rollouts import transcription_rollout
+from modal_training_gym.train_recipes.slime_recipe.recipe import SlimeRecipe
+
+
+@dataclass(config=ConfigDict(extra="forbid", arbitrary_types_allowed=True))
+class Qwen3ASR_Recipe(SlimeRecipe):
```
joyliu-q: Rename to Qwen3_ASR_1_7B_Recipe

modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -480,7 +490,12 @@ def get_base_recipe(cls, model_config: ModelConfig) -> "SlimeRecipe | None":
         from modal_training_gym.train_recipes.slime_recipe.qwen3_6_35b import (
             Qwen3_6_35b_Recipe,
         )
+        from modal_training_gym.train_recipes.slime_recipe.qwen3_asr import (
+            Qwen3ASR_Recipe,
+        )
 
+        if model_config.model_name == "Qwen/Qwen3-ASR-1.7B":
+            return Qwen3ASR_Recipe()
```
joyliu-q: Rename as mentioned above

modal_training_gym/__init__.py
```diff
@@ -10,6 +10,7 @@
     "EvalConfigDurable": ("modal_training_gym.common.eval", "EvalConfigDurable"),
     "EvalResult": ("modal_training_gym.common.eval", "EvalResult"),
     "EvalRowResult": ("modal_training_gym.common.eval", "EvalRowResult"),
+    "evaluate_asr": ("modal_training_gym.common.eval", "evaluate_asr"),
```
joyliu-q: `evaluate_asr` is too general, so no one will discover or import this. Why do we want to make this a global helper function users can use, and is there a better name we can have that explains what this fn is doing?

modal_training_gym/__init__.py
```diff
@@ -52,6 +54,15 @@
     "Qwen3_30B": ("modal_training_gym.common.models", "Qwen3_30B"),
     "Qwen3_32B": ("modal_training_gym.common.models", "Qwen3_32B"),
     "Qwen3_6_35B": ("modal_training_gym.common.models", "Qwen3_6_35B"),
+    "Qwen3ASR": ("modal_training_gym.common.models", "Qwen3ASR"),
+    "Qwen3ASR_Recipe": (
+        "modal_training_gym.train_recipes.slime_recipe",
+        "Qwen3ASR_Recipe",
+    ),
+    "LibriSpeechASRDataset": (
```
joyliu-q: This feels like an implementation Dataset tied to only the tutorial example. Can we not have it in the global library import

tutorials/rl/006_audio_asr/006_audio_asr.py
```diff
@@ -0,0 +1,133 @@
...
+# and `.label` (the reference). We score it as **negative word error rate** so
+# that lower WER → higher reward, and GRPO pushes the model toward more accurate
+# transcripts. (`jiwer` is installed for you with the model.)
+#
+# Qwen3-ASR is already near-perfect on clean LibriSpeech, so the `Qwen3ASR_Recipe`
+# defaults sample many transcripts per clip at temperature 1.0 — that's what gives
+# the GRPO group enough within-group WER variance to produce a non-zero gradient.
+
+async def wer_reward(args, sample, **kwargs) -> float:
+    import jiwer
```
joyliu-q, line 143: nit: rename to `word_error_rate_reward`? Had to look this one up.

tutorials/rl/006_audio_asr/006_audio_asr.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # ## Load LibriSpeech audio
+    #
+    # `LibriSpeechASRDataset` pulls a few clips from the standard LibriSpeech dummy
+    # set. Each row is a prompt with an `<audio>` placeholder, the clip itself (as a
+    # base64 `data:audio/wav` URI), and the reference transcript as the label. As a
+    # `MultimodalDataset` it tells slime to forward the audio column to the rollout —
+    # the same passthrough images and video use.
+
+    dataset = LibriSpeechASRDataset(n_rows=8)
```
joyliu-q, line 233: As mentioned above, this dataset should not be a global import. How will a user read tutorial and know how to define their own audio dataset

modal_training_gym/rollouts/audio_transcription.py
BenCowen: i forgot to do this split-

modal_training_gym/common/models/qwen3_asr_1_7b.py
```diff
@@ -0,0 +1,147 @@
...
+    model_name = "Qwen/Qwen3-ASR-1.7B"
+
+    # The native megatron-bridge Qwen3-ASR forward doesn't implement THD sequence
+    # packing, so training must use padded (bshd) batches; the slime launcher
+    # enforces this when the recipe leaves slime's default thd packing on.
+    requires_bshd = True
+
+    # The processor expands this single <|audio_pad|> to N tokens (N = the audio
+    # encoder's output length for the clip) so audio embeddings align with token
+    # positions. It must appear in the prompt TEXT, and the raw audio data-URI must
```
joyliu-q: Nit: This is not properly formatted
>  It must appear in the prompt TEXT, and the raw audio data-URI must NOT — inlined base64 tokenizes into ~10^5-10^6 text tokens (length scales with clip duration) and the actor OOMs.

modal_training_gym/common/models/qwen3_asr_1_7b.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Loaded from the in-image checkpoint, so the SGLang import is deferred to call
+    time and the cache key is the checkpoint path.
+    """
+    from sglang.srt.configs.qwen3_asr import Qwen3ASRProcessor
+
+    return Qwen3ASRProcessor.from_pretrained(checkpoint)
+
+
+def _prompt_user_text(prompt: Any) -> str:
```
joyliu-q: Can we add in a comment what prompt is? Seems to me that it can be a list or str object(?). It should be typed more tightly for readability.

I ended up drilling this all the way up to slime's `Sample` type, which may have a prompt input, but we also default `""` if it's not found. In the default case, `_prompt_user_text` would also return an empty string for user instruction text. Is this expected?

Will we ever get prompt as not an input?

These are some open questions but would love to understand the behavior on this model

```
getattr(sample, "prompt", "")
```

modal_training_gym/common/models/qwen3_asr_1_7b.py
```diff
@@ -0,0 +1,147 @@
...
+    user = f"{placeholder}\n{user_text}".strip() if user_text else placeholder
+    return f"<|im_start|>user\n{user}<|im_end|>\n<|im_start|>assistant\n"
+
+
+def encode_training_inputs(
+    checkpoint: str, prompt: Any, response: str, audio_bytes: bytes
+) -> dict[str, Any]:
+    """Tokenize a transcription sample for audio-conditioned GRPO training.
+
+    Returns ``{"prompt_ids", "response_ids", "multimodal_inputs"}`` — the prompt
```
joyliu-q: Since it's returning in this format, can we use a `@dataclass` to type the output and enforce this format?

modal_training_gym/common/models/qwen3_asr_1_7b.py
```diff
@@ -0,0 +1,147 @@
...
+    text = render_prompt(prompt)
+    target_sr = int(getattr(proc.feature_extractor, "sampling_rate", 16000))
+    waveform = decode_to_mono(audio_bytes, target_sr)
+
+    out = proc(text=text, audio=waveform, return_tensors="pt")
+    prompt_ids = [int(t) for t in out["input_ids"][0].tolist()]
+    response_ids = [
+        int(t) for t in tokenizer.encode(response, add_special_tokens=False)
+    ]
+    if not response_ids:  # never zero-length: prompt_length-1 must stay non-negative
```
joyliu-q: If it is never zero-length, should we error or log an error?

modal_training_gym/common/audio.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+from __future__ import annotations
+
+import base64
+from typing import Any
+
+
+def data_uri_to_bytes(data_uri: str) -> bytes:
+    """Decode a ``data:audio/<fmt>;base64,...`` URI (or a bare base64 string)."""
```
joyliu-q: Unneeded comment, self explanatory and we can remove

modal_training_gym/common/audio.py
```diff
@@ -0,0 +1,54 @@
...
+    if data_uri.startswith("data:"):
+        _, _, b64 = data_uri.partition(",")
+    else:
+        b64 = data_uri
+    return base64.b64decode(b64)
+
+
+def coerce_audio_bytes(value: Any) -> bytes | None:
```
joyliu-q: Nit: Rename to `coerce_audio_to_bytes`

modal_training_gym/common/audio.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    first = value[0] if isinstance(value, (list, tuple)) and value else value
+    if isinstance(first, (bytes, bytearray)):
+        return bytes(first)
+    if isinstance(first, str) and first:
+        return data_uri_to_bytes(first)
+    return None
+
+
+def decode_to_mono(audio_bytes: bytes, target_sr: int) -> Any:
```
joyliu-q: Is there a way we can type the return type better

modal_training_gym/frameworks/slime/modal_helpers/patches/model_specific_patches/qwen3_asr/patch_qwen3_asr_bridge_config.py
```diff
@@ -0,0 +1,39 @@
...
+
+megatron-bridge 0.5.0's hf_qwen3_asr config reads ``self.thinker_config`` in
+``get_text_config()``, but transformers 5.6's ``validate_token_ids`` calls it
+DURING ``super().__init__()`` — before ``__init__`` assigns ``thinker_config`` —
+so loading ANY Qwen3-ASR config raises AttributeError. (Same bug class sglang
+fixed in its config via PR #24187.) Guard the access; report upstream so this can
+be dropped.
```
joyliu-q, line 8: Since this is a bug that was fixed when bumping SGLang, can we add a todo comment that's

TODO(joy): Remove this patch file when bumping SGLang to past >=[version #]

modal_training_gym/common/ray_cluster.py
```diff
@@ -12,11 +12,38 @@
...
 import time
+from collections.abc import Callable
+
+from modal.experimental import clustered
 
 RAY_PORT = 6379
 RAY_DASHBOARD_PORT = 8265
 
 
+def clustered_if(
```
joyliu-q, line 32: Should we pass in the GPU type here, and RDMA always is True if the GPU type is H100/H200/B200/B300/GB200

tutorials/rl/006_audio_asr/006_audio_asr.py
```diff
@@ -0,0 +1,281 @@
...
+    return EvalRowResult(
+        score=max(0.0, 1.0 - wer),
+        response=hypothesis,
+        prompt=example["prompt"],
+        metadata={
+            "audio": audio_uri,
+            "reference": reference,
+            "wer": wer,
+            "hyp": hypothesis,
+        },
```
joyliu-q: Now that we are adding Audio as a metadata format, let's try to make an `AudioEvalRowResult(EvalRowResult)`, which is basically the same thing you added, but a user primative they can reuse.

You can use it like this
```
AudioEvalRowResult(
        score=max(0.0, 1.0 - wer),
        response=hypothesis,
        prompt=example["prompt"],
        audio=audio_uri.
        reference=reference,
        wer=wer,
        hypothesis=hypothesis,
        metadata={...}, # optional metadata can still be added here
)
```

And it directly translates to

```
EvalRowResult(
        score=max(0.0, 1.0 - wer),
        response=hypothesis,
        prompt=example["prompt"],
        metadata={
            "_metadata_type": "audio", # Note that this is a new field!!
            "audio": audio_uri,
            "reference": reference,
            "wer": wer,
            "hyp": hypothesis,
        },
)
```

Note that unlike before, you added `_metadata_type`, which will allow your frontend in `EvalsPage` to be able to auto-detect if it should render an audio cell in the evals. This also helps you and make sure you don't need to do these weird valiation since if `_metadata_type` is audio, you should *always* have the other fields such as `audio`/`reference` populated.

This also will lead to some changes in `EvalsPage.svelte`.

Note that this is a *specific* type of audio RL, so I could be wrong in which fields are optional and what other things may need to be added (text to speech? speech to text? speech to speech?). Can you respond in this thread with what you think makes the most sense for `AudioEvalRowResult` to have as fields?

joyliu-q (commented):
Almost there! Thank you again for responding to these comments.

tutorials/rl/006_audio_asr/006_audio_asr.py
BenCowen: My instinct was for something general purpose with minimal task jargon like:
```
  MultimodalEvalRow(                                                               
      score=1 - wer,                                  # headline number, drives the histogram                                                                        
      input={"audio": clip_uri, "text": instruction}, # what the model saw         
      output={"text": hypothesis},                    # what it produced           
      target={"text": reference},                     # ground truth (optional)    
      metrics={"wer": wer},                           # any other task metrics
  )                       
```

I think real datasets could be incomplete (sometimes [audio, text], sometimes [image,text,audio]). Then we could have (audio, text, image, video) widgets in the dashboard and eagerly choose which to render per eval sample.

Maybe this ought to be a follow-up PR?

BenCowen: @jvmncs just brought up a good point that the response is likely to be some structured Pydantic object that may complicate the visualization a bit

modal_training_gym/common/models/qwen3_asr_1_7b.py
BenCowen: Just making sure i follow, the em-dash?

modal_training_gym/common/models/qwen3_asr_1_7b.py
BenCowen: In this case the prompt is indeed optional, as audio alone can drive transcription (in _practice_ I suspect you would probably always have a prompt..).

I agree that string or list with default `""` makes sense here.

modal_training_gym/common/models/qwen3_asr_1_7b.py
BenCowen: For ASR, if you give the model an audio clip with no words it should return nothing (for example experimental jazz) so we wouldn't want to interrupt with a raise.

tutorials/rl/006_audio_asr/006_audio_asr.py
BenCowen: For now I went ahead and made a 
```
  AudioEvalRowResult(                                       
      score=max(0.0, 1.0 - wer),                            
      response=hypothesis,                                  
      prompt=example["prompt"],                             
      audio=audio_uri,         
      reference=reference,
      metrics={"wer": wer},     # flexible for customer to specify metrics
      metadata={...},           
  )                       
```

joyliu-q: MultimodalEvalRow makes sense as a follow-up PR, and perhaps in the follow-up PR we can make `AudioEvalRowResult(MultimodalEvalRow)`.

Changes for now makes sense!

joyliu-q (approved):

PR #111, opened 2026-06-07
Add Qwen3-VL-8B GUI grounding tutorial (rl/008_computer_use)

zhouhelena1:
Adds an RL tutorial that trains **Qwen3-VL-8B-Instruct** via GRPO to predict click coordinates from GUI screenshots. Given a screenshot and instruction like "click the Submit button," output a normalized (x, y) click point.

image
image
image

Results: Base model: 127/200 (63.5%) → Trained model: 173/200 (86.5%), Delta: +23.0%

Full wandb results here: https://wandb.ai/helenaaspamm/computer-use-grounding

### Model & Recipe

**`Qwen3_VL_8B`** (`common/models/qwen3_vl_8b.py`) — HF model configuration for the text backbone architecture (from `config.json` → `text_config`), with `requires_bshd=True` for the vision encoder.

**`Qwen3_VL_8b_Recipe`** (`train_recipes/slime_recipe/qwen3_vl_8b.py`) — VL-specific GRPO defaults:
- Frozen vision tower (`freeze_params_name_list=["vision_model"]`) — RL only updates the language backbone
- Padded (bshd) batches for the vision encoder (`use_dynamic_batch_size=False`, `qkv_format: "bshd"`)
- TP=2 colocated on 1×8×H100
- Short 256-token response cap, reduced SGLang memory fraction (0.55) for the ViT
- AutoBridge mode (`megatron_to_hf_mode="bridge"`) to skip slime's torch_dist pre-conversion

### Tutorial (`tutorials/tutorial_generator/rl/008_computer_use.py`)

- **Dataset**: `ScreenSpotDataset` subclassing `MultimodalDataset`, wrapping [rootsautomation/ScreenSpot](https://huggingface.co/datasets/rootsautomation/ScreenSpot) — ~1,200 GUI screenshots with bbox annotations. Each row keeps the full `[left, top, right, bottom]` bounding box as the label and asks the model to output a single `(x, y)` click point. Trains on 800, evals on 200.
- **Reward**: `+1` when the predicted click lands inside the target bbox, linear decay to `−1` outside over a margin equal to the element's diagonal (floored at 5% of screen for tiny targets), hard `−1` for parse failures.
- **Eval**: Binary hit-or-miss score via `ImageEvalRowResult`. A hit is when the predicted `(x, y)` falls inside the element's bounding box.

### Slime VL patches

**`patch_qwen3_vl_export.py`** — Per-param Megatron→HF converter for Qwen3-VL. Maps `language_model.*` to the standard Qwen3 decoder namespace and passes the frozen vision tower through as an identity rename (`vision_model.*` → `model.visual.*`).

**`patch_qwen3_vl_torch_dist.py`** — torch_dist→HF shim that skips the frozen ViT's stacked QKV layers (which slime's default splitter mis-assigns). Passes vision weights through unmodified.

Both are baked into the slime base image via `launcher.py`

### Other details

- `phase_reporting.py` gains per-sample input-image capture for image-modality runs:
- Show rollout input images in dashboard sample panel
- Safe async reloads on metadata for rollout display and tests for that

## Design decisions

- **Frozen ViT**: RL with sparse reward is too noisy to safely fine-tune a pretrained visual encoder — you'd risk collapsing its features. Grounding is really about teaching the decoder to read out coordinates from features the ViT already provides, and freezing is cheaper (no optimizer state or backward pass for the ViT).
- **Bridge mode for export**: slime's MB→HF converter assumes a single decoder stack and breaks on the vision tower. The workaround is freezing the ViT, using AutoBridge mode for loading, and refilling ViT weights from the base HF checkpoint at export time.

## Checklist

- [X] Example pins its dependencies
  - [X] Example pins container images to a stable tag, not a dynamic tag like `latest`
  - [X] Example specifies a `python_version` for the base image, if it is used
  - [X] Example pins all dependencies to at least minor version, `~=x.y.z` or `==x.y`
  - [X] Example dependencies with `version < 1` are pinned to patch version, `==0.y.z`


modal_training_gym/common/dataset.py
```diff
@@ -606,7 +606,7 @@ def _to_row(self, r: dict[str, Any]) -> DatasetRow:
             self.label_key: r["label"],
         }
 
-    def load(self) -> list[DatasetRow]:
+    def load(self, split: Literal["all", "train", "eval"] = "all") -> list[DatasetRow]:
```
joyliu-q: Why are we adding this here?

modal_training_gym/common/deployment.py
```diff
@@ -278,16 +278,32 @@ def _serialize_deployment_config(
             "served_model_name": value.served_model_name,
         }
 
-    def generate(self, prompt: str, ensure_ready: bool = True, **kwargs) -> str:
+    def generate(
+        self,
+        prompt: str,
+        ensure_ready: bool = True,
+        images: list[str] | None = None,
```
joyliu-q: This seems too specific of a change for a generic function, can image be passed in as comma separated value in prompts?

joyliu-q: Especially with this very specific impl detail for your specific image training case
```
            content: object = [
                {"type": "text", "text": prompt},
                *(
                    {"type": "image_url", "image_url": {"url": image}}
                    for image in images
                ),
            ]
```

modal_training_gym/train_recipes/slime_recipe/__init__.py
```diff
@@ -32,4 +33,5 @@
     "Qwen3_6_27b_Recipe",
     "Qwen3_6_35b_Recipe",
     "Qwen3_ASR_1_7b_Recipe",
+    "Qwen3VL_Recipe",
```
joyliu-q: Qwen3_VL_Recipe, not Qwen3VL_Recipe
Follow model naming convention

modal_training_gym/common/deployment.py
zhouhelena1: Got it, I've modified so that images can be passed into prompt as list[dict] since they take the shape `{"type": "text", "text": ...}`

modal_training_gym/common/dataset.py
zhouhelena1: Removed, the split was doing nothing. I had left it over to match the other DatasetConfig.load and HuggingFaceDataset.load that accept a split param

pyproject.toml
```diff
@@ -7,6 +7,7 @@ dependencies = [
     "modal==1.5.1",
     "cloudpickle",
     "datasets",
+    "pillow",
```
joyliu-q: Where is pillow used?

modal_training_gym/utils/metadata.py
```diff
@@ -127,6 +127,13 @@ def _safe_reload(vol) -> None:
         pass
 
 
+async def _safe_reload_async(vol) -> None:
+    try:
+        await vol.reload.aio()
+    except RuntimeError:
+        pass
+
```
joyliu-q, line 135: Is this fixed in other branch?

joyliu-q (approved):
LGTM minus comments with unnecessary dep?

pyproject.toml
zhouhelena1: It's to display the images in the rollout section of dashboard, in `phase_reporting.py`, but I realized it's only used in the rollout container not in the local, so I'll remove it from here.

modal_training_gym/utils/metadata.py
zhouhelena1: Yes it's fixed in main already, confirmed no diffs between metadata.py in main and in this branch. Not sure why it's still showing up in the PR.

PR #119, opened 2026-06-09
Fix megatron bridge mode

atoniolo76:
Set ref_load in launcher.py to resolved local HF snapshot dir and remove requirement to torch_dist path in bridge mode. Fix bridge mode to skip None tasks for both load and save paths in patch_bridge_none_task.py.


joyliu-q (approved):
approved given format fix

atoniolo76:
image

Validated with Qwen3.6-35b-A3b

PR #129, opened 2026-06-10
Bump Harbor version + Use Harbor without cursed resources

joyliu-q:
## Checklist

## Outside contributors


jvmncs (approved):
🥇

PR #142, opened 2026-06-12
Verify all directory snapshots are unexpired from Modal dict

atoniolo76:
Toolathlon directory snapshots are saved in a Modal dict. All entries expire after 7 days of inactivity. Verifies that reads on every snapshot key return a value.

## Checklist

## Outside contributors

### DIFF
--- a/modal_training_gym/common/environments/toolathlon.py
+++ b/modal_training_gym/common/environments/toolathlon.py
@@ -543,19 +543,25 @@ def build_snapshot_library(
 ) -> None:
     """Build the ``(task, K)`` snapshot catalog, one Modal container per task, in parallel.
 
-    Resumable: a task is skipped if its final snapshot (K = len(golden)) is already cataloged, so
-    re-runs after a partial build (or a TTL refresh of only some tasks) are cheap.
+    Resumable and self-healing: a task is skipped only if *every* snapshot K = 0..len(golden) is
+    still cataloged (catalog entries expire after 7 days of inactivity — see
+    ``DirectorySnapshotLibrary.missing_steps``, whose reads also refresh the surviving entries'
+    TTLs). Tasks with any expired/missing step are rebuilt in full.
     """
     import modal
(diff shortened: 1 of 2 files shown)


atoniolo76:
TODO(@joyliu-q / @atoniolo76) Switch to named images

modal_training_gym/common/environments/toolathlon.py
```diff
@@ -543,29 +543,35 @@ def build_snapshot_library(
...
 
     app = modal.App(f"{config.sandbox_app}-builder")
 
     @app.function(
-        image=modal.Image.debian_slim(python_version="3.12").pip_install(
-            "modal~=1.4.3"
-        ),
+        image=modal.Image.debian_slim(python_version="3.12")
+        .pip_install("modal~=1.4.3")
+        .add_local_python_source("modal_training_gym", copy=True),
```
joyliu-q, line 574: Why are we adding this

atoniolo76: The modal function calls functions outside of its scope like `build_task_snapshots`. Adds the source into the image to prevent `Module modal_training_gym` not found errors.

joyliu-q (approved):

modal_training_gym/common/environments/toolathlon.py
joyliu-q: Can we instead of adding local python source, ensure *also* that the local python source is in the image where this is ran in?

joyliu-q: You have the `modal_training_gym` folder but users won't so we should validate it's baked into the image. It very well already could be.

atoniolo76: Python seems to resolve it through `site-packages`, which is where `pip install` places the package source code. Validated on devbox in a different working directory with non-editable pip install from training-gym path.

PR #148, opened 2026-06-15
Add vime (slime + vLLM) to training-gym

BenCowen:
Mirrors @joyliu-q's [#86 (Add miles to training-gym)](https://github.com/modal-projects/training-gym/pull/86): adds [vime](https://github.com/vllm-project/vime) as a framework. Vime is the vLLM project's fork of slime — it keeps slime's Megatron training stack and Data Buffer dataflow but swaps the rollout backend from SGLang to **vLLM + vllm-router**. In the gym that's a one-line change: pick a `VimeConfig` recipe instead of a `SlimeRecipe`.

## What's here
- **`train_recipes/vime_recipe/`** — `VimeConfig` (slime CLI surface with `--sglang-*` swapped for `--vllm-*` / `--vllm-router-*` / `--router-*`; image pinned to `inferactinc/public:vime-latest`) + a `Qwen3_4b_Vime_Recipe` preset.
- **`frameworks/vime/`** — `build_vime_app` launcher + `modal_helpers` (Ray bring-up, `VIME_HOST_IP`, bridge weight-load, `train.py` Ray-job assembly).
- **Wiring** — `RecipeType.VIME`, `Framework.VIME`, `VimeStatus`, dispatch in `common/train.py`, public export of `VimeConfig`/preset, and the `list_checkpoints` VIME branch.
- **Serving** — `convert_checkpoint_to_hf` is now framework-aware: a vime checkpoint converts under the **vime image** using vime's in-image `tools/convert_torch_dist_to_hf.py` (a vime torch_dist checkpoint's pickle references `vllm`, which the slime image lacks). slime/miles keep the slime path.
- **Tests** — `tests/test_vime_recipe.py`: the vime-specific vLLM/router CLI surface and the fresh-bridge load fallback.

`VimeConfig` is discoverable via the public export.

## Validation on Modal (Qwen3-0.6B haiku GRPO, verifiable reward)

End-to-end **train → vLLM rollout → serve → eval**, on both single- and multi-node, with persisted `TrainResult`s and `EvalResult`s. (Short 1-iteration runs: these exercise the full pipeline, not model convergence.)

| Stage | Single node (1×H100) | Multi-node (2×8 H100) |
| --- | --- | --- |
| Train + vLLM rollout + reward + Megatron GRPO + weight sync + checkpoint | ✅ `distinct-slur-69d7149ad34c` | ✅ `moist-laser-63d14ecc1be9` — sharded across 16 ranks on two nodes, cross-node weight sync via vllm-router |
[...]

PR #150, opened 2026-06-15
Recipe-merge fix, Framework enum consistency, and save-path tests

BenCowen:
Small robustness fixes + tests for the training config/save path, surfaced while bringing up the Vime PR (#148) end-to-end.

> This branch started as a `Volume.reload()` fix, but #151 landed the same fix independently — so that's merged in and dropped here. What remains:

### 1. Recipe merge: presets *and* subclass recipes both win
#151 made `_merge_recipe` always copy the user recipe's value over the model preset, so a plain `SlimeRecipe`'s untouched defaults clobbered the preset (e.g. the `Qwen3-4B` preset's `n_samples_per_prompt=8` → default `2`), failing `test_recipe_param_summary`. The prior rule (override only when value ≠ default) had the opposite problem: it dropped fields a recipe *subclass* deliberately pins to a default value (e.g. the long-context recipe's `context_parallel_size=1`).

Fix: a field overrides the preset if the recipe subclass **declares it** (intentional, even when it equals the default) **or** its value differs from the default. A plain `SlimeRecipe` declares nothing extra, so it falls back to the latter and no longer clobbers presets. (Folds in @joyliu-q 's #151 rollback request.)

### 2. Framework enum consistency on TrainResult
The slime launcher set `TrainResult`'s framework to the literal `"slime"` while miles passed the `Framework` enum. `TrainResult.framework` is typed `Framework`, so unify on the enum (json-safe via #149's `(str, Enum)`).

### 3. Save-path regression tests (`tests/test_metadata_persistence.py`)
The reload + enum bugs both slipped in because nothing exercised a run through to `save()` without a full GPU train. These drive the real `save()` chain against an in-memory volume — no GPU, no Modal — and a parametrized `json.dumps` over every `Framework`. Plus an opt-in remote variant (`RUN_MODAL_TESTS=1`) that runs `save()` in a container with the metadata volume unmounted.

### DIFF
--- a/modal_training_gym/common/train.py
+++ b/modal_training_gym/common/train.py
@@ -55,11 +55,27 @@ def _stop_app(app_id: str) -> None:
 
 def _merge_recipe(base: SlimeRecipe, overrides: SlimeRecipe) -> SlimeRecipe:
     base_fields = {f.name: getattr(base, f.name) for f in _dc.fields(base)}
+
+    # Fields that a recipe *subclass* declares in its own body are intentional
+    # config and must override the model preset even when they equal the
+    # SlimeRecipe default (e.g. a long-context recipe pinning
+    # context_parallel_size=1, or disabling use_kl_loss). For a plain SlimeRecipe
+    # (no subclass layer) this set is empty, so we fall back to "value differs
+    # from default" — which keeps an untouched recipe from clobbering the preset
+    # with bare defaults (e.g. a preset's n_samples_per_prompt=8 vs default 2).
+    declared: set[str] = set()
+    for cls in type(overrides).__mro__:
+        if cls is SlimeRecipe:
+            break
+        declared |= set(getattr(cls, "__annotations__", {}))
+
     for f in _dc.fields(overrides):
         if f.name not in base_fields:
             continue
         user_val = getattr(overrides, f.name)
-        base_fields[f.name] = user_val
+        default_val = _field_default(f)
+        if f.name in declared or default_val is _dc.MISSING or user_val != default_val:
+            base_fields[f.name] = user_val
     return type(base)(**base_fields)
 
 
--- a/modal_training_gym/common/train.py
+++ b/modal_training_gym/common/train.py
@@ -58,14 +58,17 @@ def _merge_recipe(base: SlimeRecipe, overrides: SlimeRecipe) -> SlimeRecipe:
 
     # Fields that a recipe *subclass* declares in its own body are intentional
     # config and must override the model preset even when they equal the
-    # SlimeRecipe default (e.g. a long-context recipe pinning
-    # context_parallel_size=1, or disabling use_kl_loss). For a plain SlimeRecipe
-    # (no subclass layer) this set is empty, so we fall back to "value differs
-    # from default" — which keeps an untouched recipe from clobbering the preset
-    # with bare defaults (e.g. a preset's n_samples_per_prompt=8 vs default 2).
+    # framework base recipe's default (e.g. a long-context recipe pinning
+    # context_parallel_size=1, or disabling use_kl_loss). We collect those by
+    # walking the MRO from the concrete recipe down to — but not including — the
+    # framework's base recipe class (the immediate subclass of BaseTrainRecipe,
+    # e.g. SlimeRecipe / MilesConfig). For a plain base recipe (no subclass
+    # layer) this set is empty, so we fall back to "value differs from default"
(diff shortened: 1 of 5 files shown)


modal_training_gym/common/train.py
```diff
@@ -55,11 +55,27 @@ def _stop_app(app_id: str) -> None:
...
+    # config and must override the model preset even when they equal the
+    # SlimeRecipe default (e.g. a long-context recipe pinning
+    # context_parallel_size=1, or disabling use_kl_loss). For a plain SlimeRecipe
+    # (no subclass layer) this set is empty, so we fall back to "value differs
+    # from default" — which keeps an untouched recipe from clobbering the preset
+    # with bare defaults (e.g. a preset's n_samples_per_prompt=8 vs default 2).
+    declared: set[str] = set()
+    for cls in type(overrides).__mro__:
+        if cls is SlimeRecipe:
+            break
```
joyliu-q: Can we not hardcode this to Slime, but also make sure it applies to other frameworks

tests/test_metadata_persistence.py
```diff
@@ -0,0 +1,135 @@
+"""Save-path regression tests for the metadata layer.
+
+Two bugs reached `main` because nothing exercised a run all the way to
```
joyliu-q: This feels like a comment in the PR description, instead of in the test code. Rephrase to describe what the test is trying to accomplish.

tests/test_metadata_persistence.py
```diff
@@ -0,0 +1,135 @@
...
+import pytest
+
+from modal_training_gym.common import run as run_mod
+from modal_training_gym.common.framework import Framework
+from modal_training_gym.common.train_result import TrainResult
+from modal_training_gym.utils import metadata
+from modal_training_gym.utils.metadata import MetadataStore
+
+
+class _FakeVolume:
```
joyliu-q: Can we move this to a common test case that can be shared/imported

tests/test_metadata_persistence.py
```diff
@@ -0,0 +1,135 @@
...
+                return False
+
+            def put_file(self, fileobj: io.BytesIO, path: str) -> None:
+                files[path] = fileobj.read()
+
+        return _Batch()
+
+
+@pytest.fixture
+def fake_volume(monkeypatch) -> _FakeVolume:
```
joyliu-q: Ditto

tests/test_metadata_persistence.py
```diff
@@ -0,0 +1,135 @@
...
+    ).save()
+
+    blob = fake_volume.files[f"{MetadataStore.TRAINING_RUNS.value}/t1.json"]
+    assert json.loads(blob)["framework"] == "slime"
+
+
+def test_train_result_save_survives_unmounted_volume(fake_volume):
+    """TrainResult.save() is the path that hit both the reload crash and the
+    non-serializable Framework enum (asdict keeps the enum)."""
+    TrainResult(app_name="a", framework=Framework.MILES, training_run_id="t2").save()
```
joyliu-q: Should this be parametrized across frameworks as well

tests/test_metadata_persistence.py
```diff
@@ -0,0 +1,135 @@
...
+
+    blob = fake_volume.files[f"{MetadataStore.TRAIN_RESULTS.value}/t2.json"]
+    assert json.loads(blob)["framework"] == "miles"
+
+
+@pytest.mark.parametrize("fw", list(Framework))
+def test_train_result_payload_is_json_serializable(fw):
+    """Every Framework must round-trip through plain json.dumps — guards against
```
joyliu-q: No comment needed

joyliu-q (approved):

BenCowen:
Addressed the review in 63f8a7a:

- **`FakeVolume` + `fake_volume` fixture** → moved into `tests/conftest.py` so they're shared/importable instead of living in one test module.
- **`test_train_result_save_survives_unmounted_volume`** → parametrized across `list(Framework)`.
- **Module docstring + per-test docstrings** → rephrased to describe what the tests verify (present-tense behavior) rather than reading like a PR description; dropped the redundant comment on the json-serializable test.
- **`_merge_recipe`** → no longer hardcodes `SlimeRecipe` as the MRO stop point. It now breaks at the framework base recipe (the immediate `BaseTrainRecipe` subclass), so the "declared-in-subclass overrides preset" logic applies to any framework's recipe subclasses, not just Slime's.

PR #155, opened 2026-06-16
Add graceful OPD fallback

atoniolo76:
If the teacher model fails to serve a logprob response when training with OPD, Slime defaults to an all-zero tensor, which introduces bias in the reverse-KL calculation to the student logprobs, which are non-zero. This bias can accumulate over a training run and hurt performance.

Introduces a guard to the patch_advantages.py that checks if the tensor is NaN—a special value that Slime clients can specify if their teacher model fails—and returns a zero tensor for reverse KL, or the expected response in this situation.

## Checklist

- [ ] Example is documented wth comments throughout, in a [_Literate Programming_](https://en.wikipedia.org/wiki/Literate_programming) style.

## Outside contributors


joyliu-q (approved):
lets goooo

PR #160, opened 2026-06-16
Show audio trajectory in training dashboard during training process

joyliu-q:


BenCowen (approved):
looks very nice in the dash!

PR #163, opened 2026-06-17
Elide base64 blobs from slime training logs

BenCowen:
## Problem
Multimodal rollouts (e.g. the Qwen3-ASR audio demo) keep their media as a `data:...;base64,<...>` data-URI on the sample prompt — that's intentional (it has to reach the engine, and #160 reads it for the dashboard audio player). But slime logs the whole prompt, e.g.:

```
RolloutManager ... sglang_rollout.py:474 - Finish rollout: ["[{'role': 'user',
'content': [{'type': 'audio', 'audio': 'data:audio/wav;base64,UklGRqTVBAB....
```

— megabytes of base64 per rollout, drowning the logs.

## Fix
`install_base64_log_eliding()` wraps `logging.Handler.handle` so any record's formatted message has long base64 runs (64+ chars after `base64,`) replaced with `<elided>` before emit:

```
Finish rollout: [{'role': 'user', 'content': [{'type': 'audio',
'audio': 'data:audio/wav;base64,<elided>'}]}]
```

It hooks `Handler.handle` (not a logger/handler filter) because slime/ray configure logging *after* the gym's preamble runs, so a logger- or handler-level filter would miss those handlers. Wired into the existing slime preamble (`patch_rollout_status_reporting`), which already runs once per process on every slime run. Idempotent; a no-op for text-only runs (no base64 to match).

`+42 / -0`, two files.

## Validation
- Unit-level: a 55 kB base64 blob logged via both `%s`-args and f-string forms collapses to `<elided>`, no long base64 left (≈47-char line).
- The injected preamble parses, and the `ImportError` fallback stub is present.
- `tests/test_audio_asr.py` (15) pass.
- ⏳ Not yet confirmed in a live slime run — would show on the next `006_audio_asr` rollout (the `Finish rollout` line should be short).


BenCowen:
Folding the base64 log-eliding into #162 (the ASR fix micro PR) so the Qwen3-ASR fixes ship together.

PR #165, opened 2026-06-17
Improve toolathlon system prompts

atoniolo76:
## Checklist

## Outside contributors


modal_training_gym/common/environments/toolathlon.py
```diff
@@ -505,7 +505,7 @@ def build_task_snapshots(
     import modal
 
     app = modal.App.lookup(config.sandbox_app, create_if_missing=True)
-    sandbox = modal.Sandbox._experimental_create(
+    sandbox = modal.Sandbox.create(
```
joyliu-q: Yo let's use sandbox v2

atoniolo76: Updated

joyliu-q (approved):

PR #166, opened 2026-06-17
Fix sglang serve crash on Modal >=1.5.0 (proxy_regions -> routing_region)

kevinlu1248:
## Problem

Deploying the sglang serve recipe on Modal **>=1.5.0** crashes with:

```
TypeError: _App._experimental_server() got an unexpected keyword argument 'proxy_regions'
```

Modal 1.5.0 renamed the routing-region kwarg on `App._experimental_server` from `proxy_regions: list[str]` to `routing_region: str` (which Modal internally expands back to `proxy_regions=[routing_region]`).

`modal_training_gym/deploy_recipes/sglang_recipe/serve_sglang.py` still passed `proxy_regions=["us-east"]`, so the decorator now rejects the call.

## Fix

Use `routing_region="us-east"` to match the current Modal API.

Note: `serve_vllm.py` is unaffected — it uses `@modal.experimental.http_server(...)` (the flash decorator), which still accepts `proxy_regions` on both 1.4.x and 1.5.0.

### DIFF
--- a/modal_training_gym/deploy_recipes/sglang_recipe/serve_sglang.py
+++ b/modal_training_gym/deploy_recipes/sglang_recipe/serve_sglang.py
@@ -110,7 +110,7 @@ def build_sglang_serve_app(
         include_source=False,
         port=sglang_port,
         exit_grace_period=25,
-        proxy_regions=["us-east"],
+        routing_region="us-east",
         target_concurrency=8,
     )
     class SGLangEndpoint:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Reusable building blocks for Modal multi-node training examples"
 requires-python = ">=3.12"
 dependencies = [
-    "modal>=1.4.0",
+    "modal>=1.5.0",
     "cloudpickle",
     "datasets",
     "msgspec",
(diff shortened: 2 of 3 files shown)


joyliu-q:
I think in pyproject.toml, we pin modal version to 1.4.1? This is also why this change only affected notebooks, and regular training with the gym installed on uv is fine as is.

Can you also bump modal client in this PR?

kevinlu1248:
@joyliu-q done — bumped the modal client in this PR.

Small correction on the pin: `pyproject.toml` had `modal>=1.4.0` (not pinned to 1.4.1); the `uv.lock` just happened to resolve to 1.4.1, which is why installed gym runs were unaffected while the notebooks (running a newer modal) hit the error.

Changes:
- `pyproject.toml`: `modal>=1.4.0` → `modal>=1.5.0`
- Relocked via `uv lock --upgrade-package modal`:
  - modal `1.4.1 → 1.5.0`
  - synchronicity `0.12.2 → 0.12.4` (modal 1.5.0 requirement)
- `serve_sglang.py`: `proxy_regions=["us-east"]` → `routing_region="us-east"` (the 1.5.0 `_experimental_server` API)

Re: typer — modal 1.5.0 dropped its `typer` dependency, but it stays in the lock as a transitive dep of `huggingface-hub`, so nothing is removed from the environment.

PR #179, opened 2026-06-23
Handle instructing users of proxy auth token creation and handle 401 errors with detailed console logging

atoniolo76:
Add graceful error handling to deployment.py instructing user to export the MODAL_KEY and MODAL_SECRET environment variable after creating a Modal proxy auth token. Instruct the user to create this token in SKILL.md. 

## Checklist

## Outside contributors


.claude/skills/run-logged/SKILL.md
```diff
@@ -0,0 +1,71 @@
+---
+name: run-logged
+description: Launch a long-running command (training runs, servers, builds) in a tmux session with unbuffered stdout so it streams live to the pane AND is teed to a log file Claude can tail. Use when the user wants to watch a long job live while Claude also monitors it, or when a Python process's output appears "frozen" because stdout is block-buffered to a pipe.
```
joyliu-q: Why are we adding this

joyliu-q (commented):
Skill seems unrelated (but could be useful in a separate PR when less generic?)

.claude/skills/run-logged/SKILL.md
atoniolo76: Oh this is some old sauce 🇲🇲

joyliu-q (approved):

PR #180, opened 2026-06-23
Fix logging for zero rewards or dict of rewards in Slime

atoniolo76:
Guard against non-numeric rewards such as aborted samples where the reward is None or a custom reward model function that returns a dict reward (e.g. per-token logprobs from the teacher). 

## Checklist

## Outside contributors

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/180" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/frameworks/slime/modal_helpers/patches/patch_zero_std_metrics.py
+++ b/modal_training_gym/frameworks/slime/modal_helpers/patches/patch_zero_std_metrics.py
@@ -1,0 +1,73 @@
+"""Patch ``_compute_zero_std_metrics`` to ignore aborted (reward is None) samples.
+
+slime's rollout metric logging path
+(``RolloutManager.generate`` -> ``_log_rollout_data`` ->
+``compute_metrics_from_samples`` -> ``_compute_zero_std_metrics``) runs on the
+kept training batch *before* the reward post-process converts samples to train
+data.  Multi-turn / agentic rollouts can yield ``ABORTED`` samples whose
+``reward`` is still ``None``: in ``generate_and_rm`` a fan-out group where *any*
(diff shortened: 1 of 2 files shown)


modal_training_gym/frameworks/slime/modal_helpers/patches/patch_zero_std_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+
+old_is_zero_std = """\
+        rewards = [sample.get_reward_value(args) for sample in samples]
+        return len(rewards) == 0 or all(rewards[0] == r for r in rewards)"""
+
+new_is_zero_std = """\
+        rewards = [
+            r
+            for sample in samples
+            if isinstance((r := sample.get_reward_value(args)), (int, float))
```
joyliu-q, line 41: If its not an int/float, what is it? Is the value == None here (and if so, should new_is_zero_std just filter out None?)

atoniolo76:
Interim fix while upstream PR is under review: https://github.com/THUDM/slime/pull/2157

modal_training_gym/frameworks/slime/modal_helpers/patches/patch_zero_std_metrics.py
atoniolo76: There are two scenarios: 1. the reward is None because a sample was aborted (this can be common in multi-turn training) or 2. the reward is a `dict` containing the per-token teacher log-probs (these must ride `sample.reward` to reach `post_process_rewards`). The right approach should just filter out any non-numeric metrics instead of falling during the parse step.

joyliu-q (approved):
LGTM

PR #187, opened 2026-06-24
allow complete slime recipes to skip model preset merge

djdumpling:
`TrainConfig` currently always merges known-model Slime recipes with the model’s built-in preset, which works for partial overrides but can be wrong for recipes that are already complete. That forced callers with custom complete recipes to work around the merge behavior, for example by monkeypatching `SlimeRecipe.get_base_recipe` so custom fields would survive. This change adds an explicit opt-out so complete recipes can be used directly while preserving the existing preset-merge behavior by default.

Also, previous changes of context length are still live.

## Checklist

- [] Example pins its dependencies

### DIFF
--- a/modal_training_gym/common/train.py
+++ b/modal_training_gym/common/train.py
@@ -117,6 +117,20 @@ def _field_default(field: _dc.Field) -> Any:
     return _dc.MISSING
 
 
+def _resolve_slime_recipe(
+    model: ModelConfig,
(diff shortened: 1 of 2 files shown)


joyliu-q (approved):

joyliu-q:
Let's do this for now, but maybe a good idea to just make recipes presets that people can explicitly extend from instead of doing this merging I mentioned!

PR #201, opened 2026-06-30
Add substep time logging

cskitty:
Add substep timing for slime training runs.

Records wall-clock timing for each phase inside every training step — generate_rollouts, offload_rollout, compute_log_probs, optimizer_step, checkpoint_save, offload_train, weight_sync, and pre/post-step evals — and persists them on the TrainingRun record alongside the existing per-step times. This is the backend for the dashboard step-timeline UI (follow-up PR: [melody/add-step-times-to-dashboard-ui](https://github.com/modal-projects/training-gym/pull/203)).

The slime driver loop is patched (patch_rollout_status_reporting.py) to emit a generic report_step_event(phase, step_event, ...) at each phase boundary; hooks that run inside actors without a rollout_id (e.g. the Megatron log-prob hook) have their reporting moved to the driver so every event is step-tagged.

Record_step_time_event (step_timing.py) writes first-seen timestamps to the shared training-gym-step-times ModalDict under the existing key scheme ({run_id}:{step}:start|finish), extended additively with substep:{name} keys. Timestamps come from the event's own event_ts, not receipt time.

After training, aggregate_step_times (pure function, unit-tested) converts raw events into step_times ({start, end, duration_s}) and substep_times per step. Substep starts are clamped to the step window (except eval_before, which legitimately precedes it), durations telescope between consecutive starts, and optional substeps (eval_before, checkpoint_save, eval_after) are only expected on their configured intervals.

Modal dict reads and deletes are batched (128 concurrent) instead of one unbounded gather, so long runs can't stampede the Dict service.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/201" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


modal_training_gym/common/step_timing.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
+
+from collections.abc import MutableMapping
+from typing import Any
+
+EVAL_BEFORE_SUBSTEP = "evaluate_rollouts"
+EVAL_AFTER_SUBSTEP = "evaluate_rollouts_end"
```
joyliu-q: Any reason we're doing global variables here instead of an enum with all the stringified substep outcomes.

modal_training_gym/frameworks/slime/phase_reporting.py
```diff
@@ -1006,17 +1006,30 @@ def report_step_start(args: Any, rollout_id: int | None = None) -> None:
     )
 
 
-def report_weight_sync(args: Any) -> None:
+def report_weight_sync(args: Any, rollout_id: int | None = None) -> None:
```
joyliu-q: Instead of these report functions being one-off `report_phase` wrappers, can we make them a more generic function that takes in the phase it is reporting, the rollout id, args, and does a `switch` statement over an enum of `SlimeStatus`es it is reporting? Will make code more compact

scripts/validate_model_configs.py
```diff
@@ -246,6 +248,10 @@ def run_base_training_on_slime(
     model_short_name = model_config.model_name.rsplit("/", 1)[-1]
     train_recipe = SlimeRecipe.get_base_recipe(model_config)
     train_recipe.num_rollout = step_count
+    if eval_interval is not None:
+        train_recipe.eval_interval = eval_interval
+    if save_interval is not None:
+        train_recipe.save_interval = save_interval
```
joyliu-q, line 246: What is the reason for adding this code?

cskitty: if either parameter is None and passed into train_recipe as None, it'll override the defaults to not eval/save rather than maintain default eval/save points. (Also not a valid value for save_interval)

modal_training_gym/common/step_timing.py
cskitty: Fixed with enum

modal_training_gym/frameworks/slime/phase_reporting.py
cskitty: Added!

modal_training_gym/frameworks/slime/modal_helpers/patches/patch_rollout_status_reporting.py
```diff
@@ -130,44 +123,240 @@ def _step_start_replacement(match: re.Match[str]) -> str:
...
                     f"{indent}{line}",
                 ]
             )
 
         src, step_start_count = step_start_pattern.subn(_step_start_replacement, src)
 
-    weight_sync_count = 0
-    if needs_weight_sync or needs_generate_rollout:
-        weight_sync_pattern = re.compile(
+    compute_log_probs_count = 0
```
pawalt, line 135: This code is really hard to review manually (basically impossible). I think we need to bring in a goldenfile-based approach here. There are existing OSS libraries for this, but I think we should just write something by hand. We should pull in slime's `train.py` at our current revision as a `.input` file in a `testdata/` directory in the tests. And save the output from a patch at `.output`. Then when running pytest, it will verify the strings match by default, and you can pass `--rewrite` to pytest to rewrite the `.output` when changes are made.

modal_training_gym/common/run.py
```diff
@@ -86,7 +90,8 @@ class TrainingRun(BaseModel):
     completed_at: int | None = None
     updated_at: int = 0
     duration_seconds: int | None = None
-    step_times: dict[str, dict[str, int | None]] | None = None
+    step_times: dict[str, dict[str, float | None]] | None = None
+    substep_times: dict[str, dict[str, dict[str, float | None]]] | None = None
```
pawalt, line 94: What does migration from old versions of the training gym look like here? What happens with `old client -> new server` and `new client -> old server`? These look like they could break. Can we give users good hints here?

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -801,57 +894,59 @@ def convert_checkpoint(
...
-
-            step_times[f"{current_step_num}"] = {
-                "start": current_step_start_time,
-                "end": current_step_end_time,
-                "duration_s": duration,
-            }
-
-        return step_times
+        ]
+        values = await asyncio.gather(*(step_times_dict.get.aio(k) for k in keys))
```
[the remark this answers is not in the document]
pawalt: Not sure whether or not this is valid. We should try on a longer run before merging this though.

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -1168,20 +1263,22 @@ async def _set_framework_status_async(status: SlimeStatus) -> None:
...
 
             step_times_read = False
-            try:
-                latest_run_record.step_times = await write_step_times(
-                    training_run_id, slime.num_rollout
-                )
-                step_times_read = True
-            except Exception as exc:
-                print(f"Failed to read step times: {exc}")
+            if not slime.async_mode:
```
pawalt, line 1289: Do we not support substep times in async? How is this plumbed to the UI? What is hte plan for this down the line?

modal_training_gym/frameworks/slime/phase_reporting.py
```diff
@@ -230,11 +232,9 @@ def log_eval_rollout_data(
 
 
 def before_log_prob_hook(args: Any, model: Any, store_prefix: str) -> None:
```
pawalt, line 234: Do we need this hook if we're doing the patching? What does this buy us, and why was this removal made?

modal_training_gym/common/step_timing.py
```diff
@@ -0,0 +1,57 @@
...
+    OFFLOAD_ROLLOUT = SlimeStatus.OFFLOAD_ROLLOUT.value
+    COMPUTE_LOG_PROBS = SlimeStatus.COMPUTE_LOG_PROBS.value
+    OPTIMIZER_STEP = SlimeStatus.OPTIMIZER_STEP.value
+    CHECKPOINT_SAVE = SlimeStatus.CHECKPOINT_SAVE.value
+    OFFLOAD_TRAIN = SlimeStatus.OFFLOAD_TRAIN.value
+    WEIGHT_SYNC = SlimeStatus.WEIGHT_SYNC.value
+    EVAL_AFTER = f"{SlimeStatus.EVAL_ROLLOUT_LOGGING.value}_end"
+
+
+def record_step_time_event(
```
pawalt, line 22: Does this work the same as previous timings in terms of data format? As you can probably tell, I'm trying to make sure we're thinking hard about forwards and backwards compat.

modal_training_gym/frameworks/slime/phase_reporting.py
cskitty: The hook is still needed as a trampoline for user-defined custom hooks configured through SlimeFrameworkConfig.custom_megatron_before_log_prob_hook. The report_phase call was removed because phase reporting has been standardized across all steps elsewhere. Kept according to/preserving existing custom-hook behavior.

modal_training_gym/common/step_timing.py
cskitty: It works the same except changing step times to floats instead of ints; anyone on a stale install with old step times will run into an error on their first new run, but the validation error should tell them to uv sync pretty easily. Old runs should still appear the same/be backwards compatible (just missing substep times).

modal_training_gym/frameworks/slime/launcher.py
cskitty: We do not; plan is to add support in the future. The UI should simply display no step/substep times right now for async runs

modal_training_gym/common/run.py
cskitty: Will show a ValidationError for step times but will add a PR description

modal_training_gym/common/step_timing.py
pawalt: Why are old runs compatible if the data type is changing? Have you validated this against old data?

cskitty: Fixed to use int step time

modal_training_gym/frameworks/slime/modal_helpers/patches/patch_rollout_status_reporting.py
cskitty: https://github.com/modal-projects/training-gym/pull/226

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -266,6 +267,100 @@ def _preflight_wandb(wandb_cfg: WandbConfig) -> str:
     return preflight_wandb(wandb_cfg)
 
 
+def aggregate_step_times(
```
pawalt, line 270: This should have a doc-level comment explaining the algorithm. I'd handwrite this comment because the AI will produce slop likely.

pawalt (approved):
lgtm once a comment is added

pawalt (approved):
lgtm

PR #219, opened 2026-07-05
stitch: disaggregated RL on Modal Flash (miles trainer + Qwen3-30B-A3B NVFP4)

joyliu-q:
**URL:** https://github.com/modal-projects/training-gym/pull/219

## Description

Disaggregated RL on Modal: the trainer runs in the app, rollouts are served by a Modal Flash pool of SGLang replicas, and weight sync is *sparse deltas over a Volume bulletin board* instead of an NCCL broadcast — so trainer and samplers scale independently and a weight update lands in place, without draining requests.

Usage is the gym's usual one call:

```python
TrainConfig(
    model=Qwen3_30B(),
    dataset=DAPOMath(),
    recipe=Qwen3_30B_A3B_Stitch_Recipe(),   # StitchRecipe(train=..., serve=...)
).train()
```

`build_stitch_app(model=, dataset=, recipe=)` returns a plain `modal.App` with `download` / `prepare_dataset` / `prepare_checkpoints` / `train` plus the `Server` Flash class; `TrainConfig` routes `RecipeType.STITCH` through it. No `modal deploy`, no separate launch step.

## Shape

`StitchRecipe` is two halves plus run identity, so neither half re-declares the other's flags:

```python
StitchRecipe(
    train=StitchTrainConfig(...),   # subclasses MilesConfig — every miles flag, zero copies
    serve=StitchServeConfig(...),   # SglangRecipe for engine args + pool/sidecar contract
    name=..., app_tags=..., wandb=...,
)
```

Miles is the only supported trainer. Cross-half values are *derived and validated at construction*, not set twice — `train.rollout_num_gpus_per_engine = serve.gpus_per_replica`, `rollout_num_gpus=0`, `colocate=False`, W&B pushed down, delta transport/encoding/checksum matched to what the sidecar expects, and a quantized run's served base checked to be a prepared local dir. A mismatch fails locally rather than 20 minutes into a Modal run.

## Example: Qwen3-30B-A3B NVFP4-46 on B200

Port of stitch's own `cookbook/miles_disagg/configs/qwen3_30b_a3b_nvfp4_46.py` (stitch `697cda7`): NVFP4 QAT on the routed-expert GEMMs (last 7 of 48 layers BF16), 1×8 B200 trainer at TP4/EP8/SP with a CPU-offloaded optimizer, GRPO+TIS on DAPO-Math-17k, and 1–3 single-B200 replicas serving the NVFP4 baseline with `--enable-cpu-weight-cache`.

That precision contract is why there's a new `prepare_checkpoints` step. The pool boots from the baseline each sparse delta is applied against, so it must be byte-exact with the trainer's export: the step materializes BF16 masters (the bridge `ref_load`) and runs miles' `tools/convert_hf_to_nvfp4.py` under the trainer's exact `NVTE_*` env. Serving mirrors it with the matching `FLASHINFER_*` settings, so rollout logprobs track the trainer's.

Pins moved to stitch main's: stitch `697cda7`, miles image `radixark/miles:dev-202607290235` + fork `modal-projects/miles@1eb7520`, SGLang fork `modal-projects/sglang@1051a95a` (base `lmsysorg/sglang:v0.5.16`) for stage/commit weight updates, plus two Megatron patches applied to the source checkout at container start (R3 dropless dispatch; reshardable CPU-offload optimizer step).

One deliberate deviation from upstream, commented in the code: `async_mode=False`. Async bounded-lag rollouts need the publish hook to wake replicas by *deployed* app name, which a single-call ephemeral run has no lookup for — replicas would only self-sync on their poll and fall outside the lag bound. The request gate therefore runs `mode=exact, lag=0` with a 240×1s retry budget.

The registry side is wired too: `Framework.STITCH` has a validation backend (`scripts/validation_backends/stitch.py`) and a dispatch-only registry entry (`run_on_pr=False` — 8 B200 trainer plus a pool, and its baseline is a one-off conversion), so `validate_model_configs.py check -m Qwen3-30B-A3B-stitch` works without putting it in any PR matrix. `list_checkpoints` accepts stitch runs, `TrainConfig(checkpoint=...)` is rejected for stitch with an explanation rather than silently ignored (the pool's delta baseline comes from `prepare_checkpoints`), and the trainer self-heals a `prepare_inputs=False` launch.

Removed: the slime-backed `SlimeStitchTrainRecipe`, the Qwen3-4B and Moonlight-16B stitch presets/examples, and the Moonlight model config added for them.

## Test plan

Local (all green): full `pytest` suite (766 passed), `ruff check` / `ruff format --check` / `compileall`, plus focused checks that the composed recipe constructs, derives the cross-half values, allocates 8 actor GPUs with `rollout_num_gpus=0`, emits the upstream SGLang server args, and that `build_stitch_app` registers `download` / `prepare_dataset` / `prepare_checkpoints` / `train` / `Server`.

Modal, on the earlier slime-backed revision of this branch (evidence that the disagg *spine* — claim, publish, in-place apply, version-gated rollouts — works end to end):
- Qwen3-4B GSM8K, 1×8 H200 + single-H200 replicas: `v0→v3`, every replica applied in place, stage 6.6–8.9s, commit 1.5–2.2s, and a version-pinned rollout returned `weight_version_start/end: 3`.
- Moonlight-16B-A3B DAPO-Math, 2×8 H200 + 4 replicas: `v1→v5`, stage 19–43s, commit ~24s, zero errors.

Not yet run: the miles/NVFP4 revision on Modal. That smoke is 8 B200 for the trainer plus 1–3 for the pool and a one-off NVFP4 conversion. Convergence is unclaimed either way — these are weight-sync smokes, not hillclimbs.

## Checklist

Link to Devin session: https://modal.devinenterprise.com/sessions/6174f132b8694b7d98086cb638a4ef7f
Requested by: @joyliu-q


modal_training_gym/train_recipes/stitch_recipe/recipe.py
```diff
@@ -0,0 +1,218 @@
...
+        if val is True:
+            out.append(flag)
+        elif isinstance(val, dict) and key in JSON_CONFIG_FIELDS:
+            out += [flag, json.dumps(val)]
+        elif isinstance(val, list):
+            out += [flag] + [str(v) for v in val]
+        else:
+            out += [flag, str(val)]
+    return out
+
```
joyliu-q: why do we need this, can we get rid of this?

modal_training_gym/train_recipes/stitch_recipe/recipe.py
```diff
@@ -0,0 +1,218 @@
...
+                "train.update_weight_transfer_mode must be 'disk-delta', got "
+                f"{train.update_weight_transfer_mode!r}"
+            )
+        # The replicas serve — and apply deltas against — exactly what checkpoint
+        # prep built for the trainer to export from.
+        if not serve.served_checkpoint_path:
+            serve.served_checkpoint_path = train.hf_checkpoint
+        elif serve.served_checkpoint_path != train.hf_checkpoint:
+            raise ValueError(
+                "the pool's baseline must be the trainer's export baseline: "
```
joyliu-q: is this overengineering

modal_training_gym/train_recipes/stitch_recipe/recipe.py
```diff
@@ -0,0 +1,218 @@
...
+    def cli_args(
+        self,
+        *,
+        model: ModelConfig | None = None,
+        dataset: DatasetConfig | None = None,
+    ) -> list[str]:
+        """The trainer half's miles CLI argv. YAML config fields
+        (:data:`YAML_CONFIG_FIELDS`) are materialized to files by the launcher,
+        which then appends the resolved flags."""
+        return self.train.cli_args(dataset=dataset, model=model)
```
joyliu-q: why do we need this

modal_training_gym/train_recipes/stitch_recipe/recipe.py
```diff
@@ -0,0 +1,218 @@
...
+        *,
+        model: ModelConfig | None = None,
+        dataset: DatasetConfig | None = None,
+    ) -> dict[str, Any]:
+        """Plain-data miles args the trainer runs with."""
+        train = self.train
+        return {
+            "fields": self.miles_fields(model=model, dataset=dataset),
+            "async_mode": train.async_mode,
+            "miles_model_script": train.miles_model_script,
```
joyliu-q: i dont like this also typed too vaugely

modal_training_gym/train_recipes/stitch_recipe/serve.py
```diff
@@ -0,0 +1,171 @@
...
+        quantization contract (e.g. the ``FLASHINFER_NVFP4_*`` mirror of the
+        trainer's ``NVTE_NVFP4_*`` settings) lives here.
+    memory : tuple[int, int] | None
+        Host-RAM (request, limit) MiB for a replica. A CPU weight cache needs
+        roughly twice the checkpoint resident, plus staging headroom.
+    ephemeral_disk : int | None
+        Container disk MiB, for a disk-mode replica's local checkpoint copy.
+    """
+
+    sglang: SglangRecipe = field(default_factory=lambda: SglangRecipe(gpu="H200"))
```
joyliu-q: Should we not default to some sglang recipe?

modal_training_gym/train_recipes/stitch_recipe/train.py
```diff
@@ -0,0 +1,224 @@
...
+            gpus_per_node=self.actor_num_gpus_per_node,
+            rollout_num_gpus_per_engine=self.rollout_num_gpus_per_engine,
+            rollout_engines=0,
+            colocate=False,
+        )
+
+    @property
+    def megatron_pythonpath(self) -> str:
+        """Where the trainer's Ray actors find source-only ``megatron.training``."""
+        return MEGATRON_PATH
```
joyliu-q: isnt this implict?

modal_training_gym/common/models/validation.py
```diff
@@ -99,4 +100,9 @@ def find(cls, name: str) -> "_ValidationConfig":
         Moonlight_16B_A3B_Instruct,
         Framework.MILES,
     ),
+    # Disaggregated: 8 B200 trainer plus a B200 rollout pool, and its served
+    # baseline is a one-off NVFP4 conversion — dispatch-only.
+    _ValidationConfig(
+        "Qwen3-30B-A3B-stitch", Qwen3_30B, Framework.STITCH, run_on_pr=False
+    ),
```
joyliu-q: Why is run_on_pr false? it should be true since its not super expensive?

modal_training_gym/common/status.py
```diff
@@ -46,10 +46,12 @@ def resolve_framework_status(phase: str, framework: str) -> FrameworkStatus | No
...
     Returns ``None`` for a phase the framework doesn't know.
     """
-    if framework.strip().lower() not in ("miles", "slime"):
+    name = framework.strip().lower()
+    # stitch runs miles in the trainer, so it reports miles' phases.
+    if name not in ("miles", "slime", "stitch"):
         raise ValueError(f"Invalid framework string detected: {framework}")
 
-    status_enum = MilesStatus if framework.strip().lower() == "miles" else SlimeStatus
+    status_enum = SlimeStatus if name == "slime" else MilesStatus
```
joyliu-q: Why are we doing this? Shouldnt we just take stitch into account

modal_training_gym/common/train.py
```diff
@@ -472,11 +491,16 @@ def framework(self) -> Framework:
...
+        if self.recipe.recipe_type == RecipeType.STITCH:
+            return Framework.STITCH
         raise TrainingGymConfigError(
             f"Unknown recipe type: {type(self.recipe).__name__}"
         )
 
     def _initializing_status(self) -> FrameworkStatus:
+        # stitch runs miles in the trainer, so it reports miles' phases too.
+        if self.recipe.recipe_type == RecipeType.STITCH:
+            return MilesStatus.INITIALIZING
```
joyliu-q: Comment is wrong. It should be "Currently, we run miles in the Stitch trainer."

scripts/diff_impact.py
```diff
@@ -177,6 +183,17 @@ def _base_recipe_for(framework, model_config):
...
+        try:
+            # Run as a script: sys.path[0] is scripts/.
+            from validation_backends.stitch import _BASE_RECIPES
+        except ImportError:  # imported as scripts.diff_impact, e.g. by tests
+            from scripts.validation_backends.stitch import _BASE_RECIPES
+
+        # An instance, like the other branches: the caller indexes by
+        # type(recipe).__name__.
+        recipe_cls = _BASE_RECIPES.get(model_config.model_name)
+        return recipe_cls() if recipe_cls is not None else None
```
joyliu-q: Why do we need this

PR #226, opened 2026-07-07
Add goldenfile test for Slime substep patching

cskitty:
Adds a goldenfile test that verifies patch_rollout_status_reporting.py inserts its step/substep timing hooks into Slime's train.py at the expected locations. Merge after [substep time PR](https://github.com/modal-projects/training-gym/pull/201). 

Currently no golden file for train_async.py since we do not support step/substep timing in async training yet.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/226" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/frameworks/slime/modal_helpers/patches/patch_rollout_status_reporting.py
+++ b/modal_training_gym/frameworks/slime/modal_helpers/patches/patch_rollout_status_reporting.py
@@ -438,5 +438,10 @@ def _patch_file(path: Path) -> None:
     print(f"Patched {path.name} with rollout status reporting")
 
 
-_patch_file(Path("/root/slime/train.py"))
-_patch_file(Path("/root/slime/train_async.py"))
+def main() -> None:
+    _patch_file(Path("/root/slime/train.py"))
+    _patch_file(Path("/root/slime/train_async.py"))
+
+
+if __name__ == "__main__":
+    main()
--- a/scripts/fetch_slime_patch_snapshots.py
+++ b/scripts/fetch_slime_patch_snapshots.py
@@ -1,0 +1,34 @@
+"""Fetch Slime source snapshots used by the local rollout-patcher golden test.
+
+The source files exist only inside the pinned Slime image, so this maintenance
+utility uses Modal to read that image. Normal pytest collection never imports or
+runs this module. It refreshes the committed ``*.input`` fixtures; CI checks the
+resulting Git diff to detect image-source drift.
+"""
+
+from pathlib import Path
+
+import modal
+
+from modal_training_gym.frameworks.slime.launcher import SLIME_IMAGE
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+TESTDATA_DIR = REPO_ROOT / "tests" / "testdata"
+SLIME_SOURCE_PATHS = {
+    "train.py": "/root/slime/train.py",
+}
+
+app = modal.App("fetch-slime-snapshots")
+image = modal.Image.from_registry(SLIME_IMAGE).entrypoint([])
+
+
+@app.function(image=image, serialized=True)
+def read_sources() -> dict[str, str]:
+    return {name: Path(path).read_text() for name, path in SLIME_SOURCE_PATHS.items()}
+
+
+@app.local_entrypoint()
+def main() -> None:
+    TESTDATA_DIR.mkdir(exist_ok=True)
+    for name, source in read_sources.remote().items():
+        (TESTDATA_DIR / f"{name}.input").write_text(source)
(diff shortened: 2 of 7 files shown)


modal_training_gym/frameworks/slime/modal_helpers/patches/patch_rollout_status_reporting.py
```diff
@@ -37,7 +37,12 @@
 
 
 def _patch_file(path: Path) -> None:
-    if not path.exists():
+    try:
```
pawalt: Why is this required? Seems like we should actually raise here.

tests/test_substep_times.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.fixture(scope="session")
+def slime_inputs(request) -> dict[str, str]:
+    # fetch from the Slime image only on --rewrite or first run
+    names = ["train.py"]
+    missing = [n for n in names if not (TESTDATA / f"{n}.input").exists()]
+    if request.config.getoption("--rewrite") or missing:
+        TESTDATA.mkdir(exist_ok=True)
+        with modal.enable_output(), app.run():
```
pawalt: why are you pulling in modal here?

tests/test_substep_times.py
```diff
@@ -0,0 +1,61 @@
...
+
+from modal_training_gym.frameworks.slime.modal_helpers.patches import (
+    patch_rollout_status_reporting as patcher,
+)
+
+SLIME_IMAGE = "slimerl/slime@sha256:269b44b17e3f7136447db4cdaa3bf36ef9e3169f1596af0d7180c45f2a301965"
+TESTDATA = Path(__file__).parent / "testdata"
+
+app = modal.App("test-substep-patches")
+image = modal.Image.from_registry(SLIME_IMAGE).entrypoint([])
```
pawalt: don't think these require modal init. won't this fail in ci?

tests/test_substep_times.py
```diff
@@ -0,0 +1,51 @@
...
+        work.write_text(source)
+        patcher._patch_file(work)
+        actual = work.read_text()
+
+        if rewrite_goldens:
+            golden_path.write_text(actual)
+            continue
+
+        assert golden_path.exists(), (
+            "run: uv run pytest tests/test_substep_times.py --rewrite"
```
pawalt: Be clear that it's because the file does not exist.

tests/test_substep_times.py
```diff
@@ -0,0 +1,51 @@
+"""Golden-file test for the Slime rollout-status patcher."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+with patch.object(Path, "exists", return_value=False):
```
pawalt: What is this patch required for? This reads as super weird to me.

tests/update_substep_goldens.py
```diff
@@ -0,0 +1,25 @@
...
+
+from modal_training_gym.frameworks.slime.launcher import SLIME_IMAGE
+
+TESTDATA = Path(__file__).parent / "testdata"
+SOURCES = {
+    "train.py": "/root/slime/train.py",
+}
+
+app = modal.App("update-substep-patch-goldens")
+image = modal.Image.from_registry(SLIME_IMAGE).entrypoint([])
```
pawalt: My questions about pulling in Modal as a dependency were not addressed here. If this code should just take a target source and patch it, we don't need to pull in Modal.

pawalt: It looks like this is used to write just the `.input` file. This seems like a reasonable approach to do this, but it's not clear when this file gets invoked. The pattern here should be a top-level comment. And this needs to be automated in CI somehow so that anytime the slime image gets updated, this will be re-run so we're always on `main`.

tests/test_substep_times.py
cskitty: The patch used to call train.py and train_async.py when run and this was a workaround to not call them when calling the patch, realized I should just move execution in a `if __name__ == "__main__"`,. Fixed this.

scripts/fetch_slime_patch_snapshots.py
```diff
@@ -0,0 +1,34 @@
...
+
+import modal
+
+from modal_training_gym.frameworks.slime.launcher import SLIME_IMAGE
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+TESTDATA_DIR = REPO_ROOT / "tests" / "testdata"
+SLIME_SOURCE_PATHS = {
+    "train.py": "/root/slime/train.py",
+}
```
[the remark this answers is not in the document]
pawalt: Seems valid that we should capture train_async.

pawalt (approved):
lgt

PR #233, opened 2026-07-09
Add in-process BFCL v3/v4 multi-turn environment

atoniolo76:
## Summary
Add the BFCL v3/v4 multi-turn tool-calling [dataset](https://gorilla.cs.berkeley.edu/blogs/13_bfcl_v3_multi_turn.html) as an environment.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/233" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/environments/bfcl.py
+++ b/modal_training_gym/common/environments/bfcl.py
@@ -509,7 +509,7 @@ def run_bfcl_episode(
                         "type": "function",
                         "function": {
                             "name": action.name,
-                            "arguments": json.dumps(action.arguments or {}),
+                            "arguments": action.arguments or {},
                         },
                     }
                     for call_id, action in zip(call_ids, actions)
--- a/modal_training_gym/common/environments/bfcl.py
+++ b/modal_training_gym/common/environments/bfcl.py
@@ -321,6 +321,19 @@ def build_prefix_messages(label: dict, K: int) -> list[dict]:
     return messages
 
 
+def prefix_turn_index(label: dict, K: int) -> int:
+    """Index of the latest user turn included in the step-``K`` prefix."""
+    latest_turn = -1
+    shown = 0
+    for turn_index, turn in enumerate(label.get("turns", [])):
+        latest_turn = turn_index
+        for _call in turn.get("calls", []):
+            if shown >= K:
+                return turn_index
+            shown += 1
+    return latest_turn
+
+
 def prune_prefix(messages: list[dict], max_messages: int) -> list[dict]:
     """Keep system + first user message and fill the rest with the most recent messages."""
     if max_messages <= 0:
@@ -436,6 +449,7 @@ def run_bfcl_episode(
     final_response = ""
     exit_reason = "max_turns"
     consecutive_errors = 0
+    current_turn = prefix_turn_index(label, start_step)
 
     for turn in range(max_turns):
         started_at = time.monotonic()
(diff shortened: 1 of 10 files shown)


modal_training_gym/common/environments/bfcl.py
```diff
@@ -0,0 +1,610 @@
...
+sandbox / directory-snapshot apparatus in :mod:`.toolathlon` is unnecessary here — replaying ground
+truth and grading both happen in plain Python, in milliseconds, with no container or network involved.
+
+This module is a thin, RL-rollout-friendly wrapper around the upstream ``bfcl_eval`` package
+(https://pypi.org/project/bfcl-eval/). It is *not* vendored — ``bfcl_eval`` is imported lazily (see
+:func:`_bfcl_eval`) so importing :mod:`modal_training_gym` doesn't pick up its large, mostly-unrelated
+dependency closure (sentence-transformers, faiss, several model-vendor SDKs, ...). Callers must
+``pip install bfcl-eval`` themselves (the tutorial's image/install cells do this).
+
+Three pieces, mirroring :mod:`.toolathlon`'s shape:
```
joyliu-q: Slop comment

modal_training_gym/common/environments/bfcl.py
```diff
@@ -0,0 +1,610 @@
...
+from modal_training_gym.common.dataset import DatasetConfig
+from modal_training_gym.common.environments.base import (
+    EvalVerdict,
+    Environment,
+    Observation,
+    StepResult,
+    ToolCall,
+)
+
+DEFAULT_CATEGORY = "BFCL_v4_multi_turn_base"
```
joyliu-q: We define this argument in 1 place only to use it `BfclMultiTurnConfig`, which is then only used in the file path.

Can you make this more direct?

Also Im assuming the only thing subject to change in the future is version #, which is separate from category? Should they be two separate fields?

modal_training_gym/common/environments/bfcl.py
```diff
@@ -0,0 +1,610 @@
...
+# The leaderboard ships no train/eval split (it's a benchmark, not a training set), so we carve our
+# own: the last N ids of the category are held out for eval.
+DEFAULT_EVAL_TAIL = 30
```
joyliu-q: This should not be a global constant if we expect others to import this class

modal_training_gym/common/environments/bfcl.py
```diff
@@ -0,0 +1,610 @@
...
+            if isinstance(params, dict)
+            else set()
+        )
+        sig = ", ".join(f"{k}*" if k in required else k for k in props)
+        lines.append(f"- {name}({sig}): {desc[:desc_chars]}")
+    return "\n".join(lines)
+
+
+def default_system_prompt(tool_schemas: dict) -> str:
+    return f"{DEFAULT_SYSTEM_PROMPT}\n\nAvailable tools:\n{render_tool_catalog(tool_schemas)}"
```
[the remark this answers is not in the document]
joyliu-q: This seems like a valid comment

modal_training_gym/common/environments/bfcl.py
```diff
@@ -0,0 +1,610 @@
...
+
+
+def build_prefix_messages(label: dict, K: int, *, obs_limit: int = 1500) -> list[dict]:
```
joyliu-q: obs limit seems possibly a constant that is a footgun down the line, especially since there's a default set

joyliu-q (commented):
Overall, I am surprised by the amount of code required to add BFCL. Is there a way we can pare this down to reduce code complexity, or create common structs that will make adding this and Toolathlon both smaller code changes?

modal_training_gym/common/environments/bfcl.py
atoniolo76: Yes, I'm not sure why this exists. K defines the terminal step in the trajectory of prefix messages.

atoniolo76:
Merged in https://github.com/modal-projects/training-gym/pull/246

PR #234, opened 2026-07-09
Fix dataset peeking in tutorials

anli5005:
Quick PR to fix two minor issues that were causing errors on some of the tutorials:

* ~~`agent/000_agent_sandbox` was migrated to v2 Sandboxes, causing the filesystem API to break as the Modal SDK is currently still pinned to v1.5.1. I switched `_experimental_create` back to `create` for now while we wait for v1.5.2 to be released. (There are other places where we use v2 Sandboxes, but it doesn't look like any of those require the filesystem API.)~~ v1.5.2 has now been released so #245 will make this unnecessary
* `rl/003_on_policy_distillation` uses `rows[:2]`, which returns a dictionary of lists instead of the other way around, causing the dataset peek loop to fail. Using `rows.select(range(2))`, which returns another dataset, fixes this.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/234" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/tutorials/agent/000_agent_sandbox/000_agent_sandbox.ipynb
+++ b/tutorials/agent/000_agent_sandbox/000_agent_sandbox.ipynb
@@ -247,7 +247,7 @@
    "source": [
     "sandbox_app = modal.App.lookup(\"agent-sandbox-tutorial\", create_if_missing=True)\n",
     "\n",
-    "sandbox = modal.Sandbox.create(\n",
+    "sandbox = modal.Sandbox._experimental_create(\n",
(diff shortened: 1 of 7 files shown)


joyliu-q (commented):
let's wait for @andrewhinh's PR that bumps to 1.5.2

anli5005:
Dropped the changes to the agent tutorial since #241 and #245 make this unnecessary by updating SDK to the just-released v1.5.2

atoniolo76 (approved):
LGTM

PR #235, opened 2026-07-09
Extract shaped reward from Slime when sample.reward is non-scalar.

atoniolo76:
In On-policy Distillation (OPD), the per-token teacher logprobs are attached on sample.reward instead of a normal scalar reward. 

Handle non-scalar rewards by accessing the "shaped_reward" on sample.metadata. Fixes error in Gym dashboard where non-zero rewards appear zero.

## Checklist

## Outside contributors

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/235" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/frameworks/slime/sample_extraction.py
+++ b/modal_training_gym/frameworks/slime/sample_extraction.py
@@ -15,6 +15,8 @@ import json
 import os
 from typing import Any
 
+from modal_training_gym.common.sample import Sample
+
 # Import path of the run model's response parser (a (str) -> ParsedResponse
 # callable). The launcher exports it; the recorder resolves and applies it.
 RESPONSE_PARSER_PATH_ENV = "TRAINING_GYM_RESPONSE_PARSER_PATH"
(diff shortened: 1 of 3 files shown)


modal_training_gym/frameworks/slime/sample_extraction.py
```diff
@@ -70,11 +70,44 @@ def _coerce_text(value: Any) -> str:
...
+# Metadata key for the dashboard score when ``sample.reward`` is not a float
+# yet (common with OPD: reward holds the teacher ``/generate`` payload until
+# post-process). Custom generate / RM hooks should set:
+#
+#   sample.metadata["shaped_reward"] = float(task_score)
+#
+_SHAPED_REWARD_KEY = "shaped_reward"
+
+
+def _sample_score(sample: Any, reward: Any = None) -> float:
```
joyliu-q: This is getting more complex, and as such since it doesn't technically allow any types anymore, we should not pass in type `Any`.

Same with the testing below: we are using `SimpleNamespace` which is a codesmell.

Can you either 1) import the slime Sample here (but then it would not be compatible with non-slime framework, so not good, or 2) import our own Sample class (strongly preferred)

modal_training_gym/frameworks/slime/sample_extraction.py
```diff
@@ -70,11 +72,35 @@ def _coerce_text(value: Any) -> str:
...
+#
+#   sample.metadata["shaped_reward"] = float(task_score)
+#
+# Extraction maps that onto gym ``Sample.score``; do not grow gym Sample with
+# slime/OPD fields like ``reward: dict``.
+_SHAPED_REWARD_KEY = "shaped_reward"
+
+
+def _sample_score(sample: Sample, reward: float | None = None) -> float:
+    """Resolve gym :class:`~modal_training_gym.common.sample.Sample.score`.
```
pawalt: wat is this comment

pawalt (approved):
lgtm

modal_training_gym/frameworks/slime/sample_extraction.py
atoniolo76: Maybe it was trying to embed an emoji XD

PR #236, opened 2026-07-10
Make entire area of table cells in training runs dashboard clickable

anli5005:
Previously, there were some areas of the training runs table that weren't clickable. This PR expands the clickable area to most of the table (except for the column containing the Expand and Open in Modal) buttons, so that clicking anywhere on the table navigates to the run detail page.

Before (clickable area highlighted):
Screenshot 2026-07-10 at 10 28 43

After:
Screenshot 2026-07-10 at 10 28 17


joyliu-q (approved):
LGTM

PR #240, opened 2026-07-10
Frontend for getting app logs for terminated runs

rhea24:
Terminated run log visibility:

- Wired up historical endpoint to frontend so we can see Modal app logs for terminated runs
- Support for filtering logs by time range
- Support for scrolling up/down in the logs triggering a load for older/newer pages of logs

Screenshot 2026-07-13 at 5 43 11 PM

For runs that are still in progress, the behavior/UI has not changed.


dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +532,197 @@
     logDropped = 0;
   }
 
+  // ── Historical logs ─────
+  let isRunning = $derived(runStatus === "running");
+
+  const HIST_PAGE = 500; // lines fetched per page
+  let histLines = $state([]); // [{id, task_id, line}], oldest-first
```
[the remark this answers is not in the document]
rhea24: i think this is fine because this is an intentional action by the user to keep loading in more log lines?

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +532,193 @@
...
+      histNextUntil = data.nextUntil;
+      queueMicrotask(() => {
+        if (el) el.scrollTop = el.scrollHeight - prevScrollHeight + prevScrollTop;
+      });
+    } catch (err) {
+      if (signal?.aborted) return;
+      histError = String(err?.message || err);
+    } finally {
+      if (!signal?.aborted) histLoadingOlder = false;
+    }
```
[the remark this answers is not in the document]
rhea24: the ordering is actually fine because scroll events in this gap are blocked (by the histLoadingOlder guard). even in the hypothetical worst case, it'd be one extra eager fetch, not a loop because the next loadHistOlder call would immediately set histLoadingOlder = true and block further calls until it completes.

dashboards/frontend/src/lib/api.js
```diff
@@ -373,6 +373,43 @@ export async function fetchRollout(trainingRunId, rolloutId) {
   return await res.json();
 }
 
+// Historical Modal logs for a run, served from the durable ClickHouse copy via
```
pawalt: We don't want to expose ClickHouse publicily here. Remove this note from the comment. It's more about the logs being durable from the client's POV.

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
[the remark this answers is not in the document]
pawalt: Is there a way to make the rendered window sliding?

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -1198,6 +1578,96 @@
           {/if}
         </div>
       {/if}
+      {:else}
+        <!-- Finished run: page through the durable ClickHouse copy on demand. -->
```
rhea24:
```suggestion
        <!-- Finished run: page through the durable copy on demand. -->
```

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +533,395 @@
...
+    if (row.ts_ns) return Number(row.ts_ns) / 1_000_000_000;
+    return Number(row.ts) || 0;
+  }
+
+  function histCursorBefore(row) {
+    if (!row) return null;
+    if (row.ts_ns) return (Number(row.ts_ns) - 1) / 1_000_000_000;
+    const ts = Number(row.ts) || 0;
+    return ts > 0 ? ts - 0.000000001 : null;
+  }
```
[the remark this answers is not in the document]
rhea24: hmm looking into this, i see theoretically how it could happen but it doesnt look like its happening in practice [edit: fixed]

modal_training_gym/_dashboard.py
```diff
@@ -215,23 +230,27 @@ def _parse_log_batches(batches: Iterable[api_pb2.TaskLogsBatch]) -> list[LogEntr
...
 
 
-def _compute_next_page(logs: list[LogEntry], limit: int) -> tuple[bool, float | None]:
-    """Derive ``(has_more, next_until)`` for keyset pagination."""
+def _compute_next_page(logs: list[LogEntry], limit: int) -> tuple[bool, str | None]:
+    """Derive ``(has_more, next_until_ns)`` for keyset pagination.
+
+    The cursor is returned as a decimal string of integer nanoseconds so it
+    survives the Python → JSON → JavaScript round-trip without float64
+    precision loss (a 2026 ns timestamp is ~1.78 × 10¹⁸, well above 2⁵³).
```
rhea24:
```suggestion
    precision loss.
```

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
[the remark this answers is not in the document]
rhea24: checked, this doesn't actually happen

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -1,4 +1,5 @@
 <script>
+  import { tick } from "svelte";
```
bokeefe123, line 2: Is this import needed?

dashboards/frontend/src/lib/api.js
```diff
@@ -373,6 +373,42 @@ export async function fetchRollout(trainingRunId, rolloutId) {
...
+  if (since != null && since !== "") params.set("since", String(since));
+  if (until != null && until !== "") params.set("until", String(until));
+  if (tail != null) params.set("tail", String(tail));
+  if (search) params.set("search", search);
+  const qs = params.toString();
+  const res = await fetch(
+    `${SERVER}/runs/${encodeURIComponent(trainingRunId)}/logs` + (qs ? `?${qs}` : ""),
+    { signal },
+  );
+  if (!res.ok) {
```
bokeefe123, line 415: Is there a reason to keep the .detail information? It seems like yes, but it breaks from the pattern here:
```js
export async function fetchEvals({ signal } = {}) {
  const res = await fetch(`${SERVER}/evals`, { signal });
  if (!res.ok) throw new Error(`HTTP ${res.status}`);
  const evals = await res.json();
  const seen = new Set();
  return evals.filter((e) => {
    if (seen.has(e.eval_id)) return false;
    seen.add(e.eval_id);
    return true;
  });
}
```

It seems like we might just want a shared error helper that abstracts this pattern so our error handling is consistent? Something kind of like this:
```js
async function errorFromResponse(res) {
  let detail = `HTTP ${res.status}`;
  try {
    const body = await res.json();
    if (body?.detail) detail = String(body.detail);
  } catch {
    // non-JSON body — keep status-code message
  }
  return new Error(detail);
}
// callsite: if (!res.ok) throw await errorFromResponse(res);
```

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +533,395 @@
     logDropped = 0;
   }
 
+  // ── Historical logs ─────
```
bokeefe123, line 558: Throughout this section there's some unnecessary comments that just state the same thing as the variable name.

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +533,395 @@
...
+
+  // Seed the pickers with the run's lifetime the first time we view a finished
+  // run.
+  let histPrefilledFor = null;
+  $effect(() => {
+    const id = runId;
+    const running = isRunning;
+    const startedAt = run?.started_at || run?.created_at || 0;
+    const endedAt = run?.ended_at || run?.completed_at || 0;
+    if (!id || running || histPrefilledFor === id) return;
```
bokeefe123: Could we not collapse these two if statements? Or if that's too verbose then move this if statement to be above `startedAt` so that the if statements are with their relevant declerations?

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
rhea24: yeah its being used in the function `restoreHistAnchor` to make sure that adjusting the scroll position happens after the corresponding `histLines` have been rendered

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +533,395 @@
...
+          top: rect.top,
+          scrollTop: el.scrollTop,
+        };
+      }
+    }
+    return null;
+  }
+
+  async function restoreHistAnchor(anchor) {
+    const el = histTailEl;
```
bokeefe123: nit:it makes it harder to read this code when this name gets aliased

dashboards/frontend/src/lib/api.js
rhea24: fixed!

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -531,6 +533,395 @@
...
+      await restoreHistAnchor(anchor);
+    } catch (err) {
+      if (signal?.aborted) return;
+      histError = String(err?.message || err);
+    } finally {
+      if (!signal?.aborted) histLoadingOlder = false;
+    }
+  }
+
+  async function loadHistNewer() {
```
bokeefe123: nit: this seems to have a lot of duplicated code with loadHistOlder. Maybe pull the duplicated code into its own helper function?

bokeefe123 (commented):
No correctness issues that I can find but there's some cleanup I'd suggest.

dashboards/frontend/src/lib/api.js
```diff
@@ -1,5 +1,16 @@
 const SERVER = "/api";
 
+async function errorFromResponse(res) {
+  let detail = `HTTP ${res.status}`;
+  try {
+    const body = await res.json();
+    if (body?.detail) detail = String(body.detail);
+  } catch {
+    // Non-JSON error body — keep the status-code message.
+  }
```
joyliu-q: no generic catch here, what exception do we expect here?

dashboards/frontend/src/lib/api.js
```diff
@@ -1,5 +1,16 @@
...
+  return new Error(detail);
```
joyliu-q: Throwing a broad Error seems like an antipattern. Also, isn't this function just formatting the error string? Better naming is `getErrorFromResponse(res) -> string` <- make sure to type things and wrap the other in `new SpecificErrorType()`

dashboards/frontend/src/lib/api.js
```diff
@@ -375,6 +386,33 @@ export async function fetchRollout(trainingRunId, rolloutId) {
   return await res.json();
 }
 
+// Historical Modal logs for a run, served from the durable storage.
```
joyliu-q, line 394: Can you explain what the durable storage is? Also what's the purpose of "historical" here?

dashboards/frontend/src/lib/api.js
```diff
@@ -375,6 +386,33 @@ export async function fetchRollout(trainingRunId, rolloutId) {
...
+//
+// Returns the newest `tail` lines within the (since, until] window, oldest
```
joyliu-q: Can we type this function

joyliu-q: This has already been merged but I do not like `tail` compared to `max_lines` or something, and the server side default is not clear here, we should comment what the default tail is from server, as well as since and until

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -548,6 +550,391 @@
     logDropped = 0;
   }
 
+  // ── Historical logs ─────
+  let isRunning = $derived(runStatus === "running");
+
+  const HIST_PAGE = 500;
+  const HIST_BUFFER_MAX = 2000;
```
joyliu-q: Global constants should be declared at the top with comments on what the constant represents

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -548,6 +550,391 @@
...
+  let histTailEl = $state(null);
+  let histSeq = 0;
+  let histController = null;
+  let histLastScrollTop = 0;
+  let histRestoringScroll = false;
+
+  let histSinceText = $state("");
+  let histUntilText = $state("");
+  let histSince = $state("");
+  let histUntil = $state("");
```
joyliu-q: There is a more concise way to represent these

dashboards/frontend/src/lib/api.js
rhea24: It's clickhouse durable storage (whereas the live streaming logs path fetches from redis but those expire after 2 hours I believe) - I had originally included this in the comment but peyton said we shouldn't expose those internal specifics here. And the 'historical' is just to distinguish from the live logs streaming path. I can get rid of it if you think it's unnecessary?

dashboards/frontend/src/lib/api.js
rhea24: quick question - what do you mean by typing the function? the file is in js and i noticed no other functions have types so i'm a bit unsure on what convention for typing is here. do you mean to add jsdoc `@typedef`s and `@param`s?

dashboards/frontend/src/lib/api.js
rhea24: fixed

dashboards/frontend/src/lib/api.js
rhea24: updated this but js has no HttpError type or something that would be more applicable here and i saw other functions in the file throw broad Errors so i kept it that way for now. i could create a new error type and throw that if you think that'd be better?

dashboards/frontend/src/lib/api.js
joyliu-q: Yeah lets specify `clickhouse durable storage` and get rid of `historical`

dashboards/frontend/src/lib/api.js
joyliu-q: ok this is fine!

joyliu-q (approved):

PR #250, opened 2026-07-14
Fix non-colocated substep timings

cskitty:
Make offload rollout and train optional for when noncolocated training runs do not include them. Currently noncolocated runs will mark substep duration of rollouts as none since the start time of the next step (offloading) does not exist, when it should just take the start time of the following step.

Also update tests and validate model configs flags accordingly to test.


pawalt (approved):
lgmt

PR #253, opened 2026-07-15
add orphaned deployment reconciler

andrewhinh:
similar to #193.


joyliu-q (commented):
Thank you! Can we maybe turn it into a singular reconciller function called `reconcille` that cleans up everything?

joyliu-q (approved):

PR #254, opened 2026-07-15
Output perf deltas when validating models in CI

anli5005:
Adds a step to our model validation CI that looks for the last successful run of each model on a merged PR. If it finds one, the CI will output the difference in training time to better highlight regressions and improvements.


scripts/download_perf_baseline.py
```diff
@@ -0,0 +1,174 @@
...
+        with _fetch_artifact_zip(artifact, token) as archive:
+            result = json.loads(archive.read(f"{artifact_name}.json"))
+            if not result.get("succeeded"):
+                continue
+            archive.extractall(baseline_dir)
+            _write_baseline_meta(
+                repo, artifact, baseline_dir / f"{artifact_name}.meta.json"
+            )
+        print(f"downloaded baseline {artifact_name!r} (artifact id {artifact.id})")
+        return True
```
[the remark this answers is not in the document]
anli5005: This is not really a problem - I think we want to be loud about failures on GitHub's end as opposed to run failures themselves.

.github/workflows/validate-models.yml
```diff
@@ -134,8 +135,17 @@ jobs:
           pattern: validate-result-*
           merge-multiple: true
 
+      - name: Download baselines from main
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: uv run python scripts/download_perf_baseline.py --baseline-dir baseline
```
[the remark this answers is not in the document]
anli5005: If `download_perf_baseline.py` itself is hosed it's probably worth being loud about it

scripts/download_perf_baseline.py
```diff
@@ -0,0 +1,176 @@
...
+        if artifact.expired:
+            continue
+        if inspected >= MAX_CANDIDATES_PER_MODEL:
+            print(
+                f"warning: gave up after inspecting {inspected} "
+                f"{artifact_name!r} candidates"
+            )
+            break
+        inspected += 1
+        if not _is_from_merged_pr(repo, artifact, merged_sha_cache):
```
joyliu-q, line 104: Thoughts on actually having a "synmon" job that runs the gym and outputs baseline artifact once every week so we don't have to do this?

joyliu-q (commented):
All the changes minus baseline detection lgtm

scripts/download_perf_baseline.py
anli5005: Piggybacking on #306 would be the best long-term solution but I think it should be fine to use something like this logic in the interim while we wait for the first few runs on that

joyliu-q: Can you add a `TODO(anthony)`

joyliu-q (approved):

PR #255, opened 2026-07-16
Remove stable identifiers for training configs

anli5005:
This PR removes `TrainConfig.training_run_id` as a stable identifier on the `TrainConfig`, migrating it to a model where IDs are generated per-run. In other words, every call to `.launch()` and `.train()` will now receive its own ID even if called on the same config instance, and will appear as separate runs on the training gym dashboard.

A few reasons why I'd like to make this change:
* It feels cleaner to keep configs as purely declarative and defer ID creation until training time
* During one of the tutorials, I called `train()` without recreating the `TrainConfig`. Doing so caused the dashboard to enter a weird state, where the old run displayed as completed but was otherwise overwritten by the details of the currently running new run.

This does mean that clients can no longer depend on reading the IDs of training runs prior to starting them. We still return the run IDs on the training run/result objects themselves, so aside from a few code changes required to use cases like param sweeps this should have little impact on functionality.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/255" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


joyliu-q (commented):
LGTM

PR #256, opened 2026-07-17
Add dashboard previews for PRs

anli5005:
Builds on the work done in #251 to build and generate previews of the dashboard. The mechanism is identical, only this time there's a new CI workflow and some nginx-based logic to proxy /api to https://modal-labs-training-gym--training-gym-dashboard-fastapi-app.modal.run/, which is the [same dashboard](https://github.com/modal-projects/training-gym/actions/runs/29529946784/job/87727657806) used in our model validation workflow.

We might change this in the future to point the API to mock data, but for now this should provide the frontend realistic data without manual intervention.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/256" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


.github/workflows/dashboard-preview.yml
```diff
@@ -0,0 +1,99 @@
...
+        run: tar -czf dashboard.tar.gz -C dashboards/frontend/dist .
+
+      - name: Upload dashboard artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: dashboard-dist
+          path: dashboard.tar.gz
+
+  preview:
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
bokeefe123, line 47: Is there any case where the `on:` list would grow and we'd want to have this if be same as your one on the documentation preview?
```yml
if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
```

anli5005: I don't see this happening so long as the preview is scoped to being frontend-only - I think there's a chance that a PR introduces both a frontend and an API change but unless we spin up a new instance of the training dashboard that's not something we can easily preview

bokeefe123 (approved):
pending security process lgtm

PR #264, opened 2026-07-22
Tiny bug fix - Qwen3_6_35b_Recipe_Long_Context so it can be used

zhouhelena1:
`Qwen3_6_35b_Recipe_Long_Context` exists in `slime_recipe/qwen3_6_35b_long_context.py` but isn't exported anywhere, so it’s usuable (unless you directly import the file). So I exported it like the other models.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/264" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


joyliu-q (dismissed):

joyliu-q:
I dont think we should add this because it's not fully tuned and I only want to add this model if it would be used in a long context training tutorial.

PR #267, opened 2026-07-22
CLI Foundations

rhea24:
Previously, `training-gym` was a single argparse dispatcher embedded in `modal_training_gym/cli/__init__.py`. It supported the existing setup and cleanup commands, but it did not provide a clean way to add nested command groups, share dashboard HTTP behavior, standardize machine-readable output, or expose stable exit codes.

This PR establishes that foundation without adding any run-inspection commands yet.

The CLI now has:
- A Click-based root command and modular command registration
- Modal-style Rich help and error rendering (all almost entirely copied over from Modal cli code, especially `commands.py`)
- Lazy wrappers preserving all existing command behavior
- A shared authenticated dashboard HTTP client
- Typed CLI errors with stable process exit codes
- Shared JSON, table, confirmation, and option helpers

Testing:
`tests/test_cli_main.py`: covers the Click command tree and process wrapper
`tests/test_cli_client.py`: covers dashboard HTTP transport behavior
`tests/test_cli_output.py`: covers shared presentation and option behavior

Screenshots:
Most of these are helpers that will be used in the future commands, but right now this is what the --help looks like (based on modal cli design)
Screenshot 2026-07-23 at 12 10 11 PM

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/267" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


modal_training_gym/cli/builtin.py
```diff
@@ -0,0 +1,66 @@
...
+def open_command() -> None:
+    """Open the deployed dashboard in your browser."""
+    from .setup import open_dashboard
+
+    open_dashboard()
+
+
+@click.command("set-proxy-auth", cls=TrainingGymCommand)
+def set_proxy_auth_command() -> None:
+    """Set/replace the Modal proxy-auth tokens (MODAL_KEY / MODAL_SECRET)"""
```
joyliu-q: Nit: comment needs to describe why/when we would need to do this

Also is MODAL_KEY the canonical name for this?

modal_training_gym/cli/client.py
```diff
@@ -0,0 +1,124 @@
...
+QueryParams = Mapping[str, str | int | float | bool | None]
+
+
+class DashboardClient:
+    """Transport shared by dashboard-backed commands."""
+
+    def __init__(
+        self,
+        *,
+        base_url: str | None = None,
```
joyliu-q: When would base_url be set? I only see this being used in testing

modal_training_gym/cli/client.py
```diff
@@ -0,0 +1,124 @@
...
+        password: str | None = None,
+        timeout: float = DEFAULT_TIMEOUT_SECONDS,
```
joyliu-q: When would we set the timeout to be a different value?

modal_training_gym/cli/client.py
```diff
@@ -0,0 +1,124 @@
...
+        transport: httpx.BaseTransport | None = None,
```
joyliu-q: Ditto

modal_training_gym/cli/client.py
joyliu-q: You can mock the result of `get_dashboard_url()`

modal_training_gym/cli/commands.py
```diff
@@ -0,0 +1,321 @@
...
+    table_width: int,
+) -> Table:
+    table = build_command_table(name_width, table_width)
+    for name, command in items:
+        table.add_row(name, command.get_short_help_str(limit=80))
+    return table
+
+
+def _available_width(console: Console) -> int:
+    return min(_MAX_HELP_WIDTH, console.width) - _HELP_PADDING * 2
```
joyliu-q: Only used once

modal_training_gym/cli/commands.py
```diff
@@ -0,0 +1,321 @@
...
+            parts.append(Text(""))
+        parts.append(section)
+    if parts:
+        parts.append(Text(""))
+    with console.capture() as capture:
+        console.print(Padding(Group(*parts), (0, _HELP_PADDING)))
+    formatter.write(capture.get())
+
+
+class TrainingGymCommand(click.Command):
```
joyliu-q: These classes names are confusing because they sound like an actual command associated with the gym library. Would recommend prefixing with `_TrainingGymCommand`.

modal_training_gym/cli/commands.py
```diff
@@ -0,0 +1,321 @@
...
+                _build_usage(self, ctx),
+                _build_help_text(self),
+                _build_options(self, ctx),
+                _build_epilog(self),
+            ],
+            formatter,
+        )
+
+
+class TrainingGymGroup(click.Group):
```
joyliu-q: These classes names are confusing because they sound like an actual group of training runs. Would recommend prefixing with `_TrainingGymGroup`.

modal_training_gym/cli/errors.py
```diff
@@ -0,0 +1,59 @@
...
+from __future__ import annotations
+
+from enum import IntEnum
+
+import click
+
+
+class ExitCode(IntEnum):
+    SUCCESS = 0
+    UNEXPECTED = 1
```
joyliu-q: What is unexpected?

modal_training_gym/cli/errors.py
```diff
@@ -0,0 +1,59 @@
...
+    USAGE = 2
```
joyliu-q: Usage??

modal_training_gym/cli/errors.py
```diff
@@ -0,0 +1,59 @@
...
+    NOT_FOUND = 3
+    AUTH = 4
+    DASHBOARD = 5
+    INTERRUPTED = 130
```
joyliu-q: These exit codes don't make sense or need to be better described. Typically 0=Good and 1=Error so these are strange

modal_training_gym/cli/errors.py
```diff
@@ -0,0 +1,59 @@
...
+    """Local dashboard configuration is missing or invalid."""
+
+
+class DashboardNetworkError(DashboardError):
+    """The dashboard could not be reached."""
+
+
+class DashboardTimeoutError(DashboardError):
+    """The dashboard request exceeded its timeout."""
+
```
joyliu-q: If the behavior under the hood is that these are just normal errors, we dont need this many classes

modal_training_gym/cli/options.py
```diff
@@ -0,0 +1,45 @@
...
+from __future__ import annotations
+
+import sys
+from collections.abc import Callable
+from typing import Any, TypeVar
+
+import click
+
+
+CommandFunction = TypeVar("CommandFunction", bound=Callable[..., Any])
```
joyliu-q, line 12: Why do we have this

modal_training_gym/cli/options.py
```diff
@@ -0,0 +1,45 @@
...
+        "--json",
+        "json_output",
+        is_flag=True,
+        default=False,
+        help="Output machine-readable JSON.",
+    )(function)
+
+
+def yes_option(function: CommandFunction) -> CommandFunction:
+    """Add the standard non-interactive confirmation flag."""
```
joyliu-q: Don't need this comment

modal_training_gym/cli/options.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Any, TypeVar
+
+import click
+
+
+CommandFunction = TypeVar("CommandFunction", bound=Callable[..., Any])
+
+
+def json_option(function: CommandFunction) -> CommandFunction:
+    """Add the standard machine-readable JSON output flag."""
```
joyliu-q: Don't need this comment

modal_training_gym/cli/output.py
```diff
@@ -0,0 +1,85 @@
...
+from collections.abc import Sequence
+from typing import Any
+
+from rich.console import Console
+from rich.panel import Panel
+from rich.table import Column, Table
+from rich.text import Text
+
+
+RICH_OUTPUT_ENV = "TRAINING_GYM_RICH_CLI"
```
joyliu-q: Doesn't feel like this is ever overridden aside from in tests, is this something we plan to do in the future?

Makes more sense to default have this everywhere and add this flag back when we have the json option (YAGNI) so the code is cleaner.

joyliu-q: Also the naming is not clear what =0 =1 is. `DISPLAY_RICH_CLI=1` vs `DISPLAY_RICH_CLI=0` is clearer

modal_training_gym/cli/output.py
```diff
@@ -0,0 +1,85 @@
...
+        width=width,
+    )
+
+
+def print_json(value: Any) -> None:
+    """Write JSON to stdout (without Rich formatting)."""
+    print(
+        json.dumps(value, ensure_ascii=False, indent=2),
+        file=sys.stdout,
+    )
```
joyliu-q: Not needed as a sep function

modal_training_gym/cli/output.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def print_table(
+    columns: Sequence[Column | str],
+    rows: Sequence[Sequence[object]],
+    *,
+    title: str = "",
+) -> None:
+    """Render a compact human-readable table to stdout."""
```
joyliu-q: Unneeded comment

modal_training_gym/cli/client.py
rhea24: og had thought there could be diff requests that take longer amounts of time (run trace, run logs) but i actually think this is fine because this timeout is for most extreme case so will get rid of this

modal_training_gym/cli/errors.py
rhea24: fixed + refactored erroring pattern so that we have exit code values (it tends to be better for agents to branch off of specific exit codes) with commented descriptions of each, and one CLIError type that has an optional recovery hint (because they are also often helpful for agents) along with message, exit code, and other details

modal_training_gym/cli/output.py
rhea24: wait i feel like it's easier to have this so that when we're branching based on the `-j` flag, it's either `print_table` or `print_json` instead of re-writing this for every command when we're determining output

modal_training_gym/cli/options.py
rhea24: it's for typing so that for `json_option` and `yes_option` both return the same callable type they receive but i could just use the type `Callable[..., Any]` directly for the inputs/outputs of those two options if you think thats better? i dont feel very strongly here

modal_training_gym/cli/client.py
```diff
@@ -0,0 +1,151 @@
...
+            httpx.BasicAuth("training-gym", dashboard_password)
+            if dashboard_password
+            else None
+        )
+        self._client = httpx.Client(
+            base_url=configured_url.rstrip("/") + "/",
+            auth=auth,
+            timeout=DEFAULT_TIMEOUT_SECONDS,
+            follow_redirects=True,
+        )
```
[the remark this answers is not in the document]
rhea24: not true - httpx gets rid of the authorization if theres a cross-origin redirect (added a test)

modal_training_gym/cli/options.py
joyliu-q: Got it, that's fine

joyliu-q (approved):
Approved provided that the change has been tested with updates

PR #269, opened 2026-07-23
Support run list cli command

rhea24:
This PR adds support for `training-gym run list` and establishes `RunSummary` as the shared source of truth for run-list data and presentation.

Previously, the dashboard derived user-facing status and stage values in JavaScript, which duplicated backend logic and made it difficult to guarantee that the CLI and dashboard presented runs consistently.

The CLI now supports:

- `training-gym run list` with a human-readable table by default
- Machine-readable output through `--json`
- Filtering by status, model, dataset, recipe, and group ID
- Time filtering through relative durations, ISO 8601 timestamps, or epoch seconds
- Result limiting and sorting by most recent update
- Schema-driven columns and filter options, allowing future fields to be exposed by annotating the corresponding `RunSummary` field

The shared run representation now includes:
- `display_status`, which normalizes raw run state into pending, completed, failed, stopped, or cancelled
- `display_stage`, which provides a human-readable framework stage and distinguishes queued GPU work from active work
- Field metadata describing which values belong in run lists, which are filterable, and which require timestamp formatting

The dashboard API now:
- Accepts the same filters used by the CLI
- Supports since and limit
- Filters and sorts through shared Python logic
- The dashboard frontend now consumes `display_status` and `display_stage` directly. This removes the duplicate JavaScript status/stage normalization and keeps dashboard presentation aligned with the CLI.

Time parsing was also moved into a shared helper so run-list filtering and historical log filtering use the same rules.

Screenshots:
Screenshot 2026-07-27 at 2 48 00 PM
Screenshot 2026-07-27 at 3 33 26 PM
Screenshot 2026-07-27 at 2 46 28 PM
Screenshot 2026-07-27 at 3 40 45 PM

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/269" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


dashboards/frontend/src/pages/DeploymentsPage.svelte
```diff
@@ -116,7 +116,7 @@
       return "Inactive";
     }
     if (endpointCreationTimedOut) return "Inactive";
-    if (row.run && getStatus(row.run) === "Failed") return "Inactive";
+    if (row.run && getStatus(row.run) === "failed") return "Inactive";
```
joyliu-q, line 119: should getStatus just be a function that returns an enum? The whole capitalization is a bit odd

modal_training_gym/common/time.py
```diff
@@ -0,0 +1,45 @@
+"""Shared parsing helpers for user-supplied time bounds."""
```
joyliu-q, line 1: Rename to time.py

modal_training_gym/common/time.py
```diff
@@ -0,0 +1,45 @@
...
+_RELATIVE_SECONDS = {
+    "s": 1,
+    "m": 60,
+    "h": 60 * 60,
+    "d": 24 * 60 * 60,
+    "w": 7 * 24 * 60 * 60,
+}
+
+
+def parse_time(value: str, now: float) -> float | None:
```
joyliu-q, line 20: Does `datetime` or another library already do this?

modal_training_gym/cli/run.py
```diff
@@ -0,0 +1,201 @@
...
+
+def parse_since(value: str, *, now: float | None = None) -> int:
+    """Parse epoch, ISO 8601, or relative time into epoch seconds."""
+    parsed = parse_time(value, time.time() if now is None else now)
+    if parsed is None:
+        raise click.BadParameter(
+            "Must be epoch seconds, ISO 8601, or a relative time such as 24h"
+        )
+    return int(parsed)
+
```
joyliu-q: Seems like a function that's only used once. Can we inline this?

dashboards/frontend/src/pages/DeploymentsPage.svelte
rhea24: i feel like since we're using js and there aren't native enums in js, it's fine to do this (despite agreeing yeah its odd)? i made it so that `display_status` is normalized by the backend to a fixed set of lowercase values, and the frontend consistently uses those lowercase values for comparisons. and then only the user-facing labels in the frontend handle the capitialization

modal_training_gym/common/time.py
rhea24: `datetime` doesn't handle parsing relative values like the 24h or 7d and i think adding a third party library might be unnecessary

modal_training_gym/cli/run.py
```diff
@@ -0,0 +1,196 @@
...
+        .isoformat(timespec="seconds")
+        .replace("+00:00", "Z")
+    )
+
+
+def _format_table_timestamp(value: object, *, now: float | None = None) -> str:
+    if not isinstance(value, (int, float)) or not value:
+        return "—"
+    age = (time.time() if now is None else now) - value
+    if 0 <= age < 60:
```
bokeefe123: nit: Clock skew could in theory allow all these statements to pass by and a raw format to go through. Aka age could be negative

modal_training_gym/cli/run.py
```diff
@@ -0,0 +1,196 @@
...
+        return "—"
+    return (
+        datetime.fromtimestamp(value, tz=UTC)
+        .isoformat(timespec="seconds")
+        .replace("+00:00", "Z")
+    )
+
+
+def _format_table_timestamp(value: object, *, now: float | None = None) -> str:
+    if not isinstance(value, (int, float)) or not value:
```
bokeefe123, line 60: funny enough a bool could pass this if statement. Don't actually need to fix because all callers handle this but it is *in theory* possible

modal_training_gym/cli/run.py
```diff
@@ -0,0 +1,196 @@
...
+
+from .client import DashboardClient
+from .commands import _TrainingGymGroup
+from .errors import CLIError, ExitCode
+from .options import json_option
+from .output import print_json, print_table
+
+
+DEFAULT_RUN_LIMIT = 50
+CLI_FIELD_NAMES = {
```
bokeefe123, line 25: If a future list field aliases to an existing output name (e.g. status), the JSON dict comprehension and the generated Click flags will silently collide. Worth an assertion/test that names are unique after applying CLI_FIELD_NAMES

modal_training_gym/common/time.py
bokeefe123, line 20: This is pretty hard to read, also it's probably good to compile the regex before every call. Here's what my Devin cooked up:
```py
_RELATIVE_RE = re.compile(r"(\d+)\s*([smhdw])", re.IGNORECASE)
 
 
def parse_time(value: str, now: float) -> float | None:
    """Parse relative age, epoch seconds, or ISO 8601 into epoch seconds."""
    text = (value or "").strip()
    if not text:
        return None
    for parser in (_parse_relative, _parse_epoch, _parse_iso):
        result = parser(text, now)
        if result is not None:
            return result
    return None
 
 
def _parse_relative(text: str, now: float) -> float | None:
    match = _RELATIVE_RE.fullmatch(text)
    if not match:
        return None
    amount, unit = int(match.group(1)), match.group(2).lower()
    return now - amount * _RELATIVE_SECONDS[unit]
 
 
def _parse_epoch(text: str, _now: float) -> float | None:
    try:
        parsed = float(text)
    except ValueError:
        return None
    return parsed if math.isfinite(parsed) else None
 
 
def _parse_iso(text: str, _now: float) -> float | None:
    try:
        parsed = datetime.fromisoformat(text)  # handles "Z" natively on 3.11+
    except ValueError:
        return None
    if parsed.tzinfo is None:
        parsed = parsed.replace(tzinfo=UTC)
    return parsed.timestamp()
```

modal_training_gym/common/time.py
```diff
@@ -0,0 +1,45 @@
...
+
+_RELATIVE_SECONDS = {
+    "s": 1,
+    "m": 60,
+    "h": 60 * 60,
+    "d": 24 * 60 * 60,
+    "w": 7 * 24 * 60 * 60,
+}
+
+
```
bokeefe123, line 19: I'd add a test for this as well.

bokeefe123 (changes requested):

bokeefe123 (approved):
after solving merge conflict lgtm

modal_training_gym/cli/run.py
```diff
@@ -0,0 +1,196 @@
...
+    }
+    with DashboardClient() as client:
+        payload = client.get_json("/api/runs", params=params)
+
+    if not isinstance(payload, list):
+        raise CLIError(
+            "Dashboard returned an invalid run list.",
+            error="invalid_dashboard_response",
+            exit_code=ExitCode.BACKEND,
+        )
```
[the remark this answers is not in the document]
rhea24: not applicable; we need to run `training-gym setup` regardless

PR #270, opened 2026-07-23
Make modal.exception import in utils.metadata import-safe

zhouhelena1:
## Summary

Reverted `utils/metadata.py` to function-local imports. Currently, every slime training run on main breaks at first rollout with `ModuleNotFoundError: No module named 'modal'`.


joyliu-q (approved):
Thanks for the find.
We should run CI more aggressively or find a cheaper way to prune these errors.

PR #280, opened 2026-07-28
Run get cli command

rhea24:
This PR adds support for `training-gym run get RUN_ID`.

The dashboard also now exposes `GET /api/runs/{training_run_id}`, which loads the run and optional train result directly without scanning the entire list-summary stores (which was the only way to access a particular run previously).

Screenshots:
Screenshot 2026-07-28 at 4 49 21 PM (1)
Screenshot 2026-07-28 at 4 55 53 PM
Screenshot 2026-07-28 at 5 15 54 PM
Screenshot 2026-07-27 at 6 25 16 PM

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/280" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


modal_training_gym/_dashboard.py
```diff
@@ -736,6 +737,25 @@ async def runs(
         )
         return filtered
 
+    @web.get("/api/runs/{training_run_id}", response_model=RunSummary)
+    async def get_run(training_run_id: str):
```
joyliu-q, line 744: Currently, this is new code and not using the shared path of how TrainingRunDetailsPage and how it fetches a specific run by run id.

Let's make the training details page also use your written get_run code:

Right now api.js has fetchRunRollouts/fetchRollout/fetchRunLogs but no fetchRun(id)

modal_training_gym/cli/run.py
```diff
@@ -98,6 +99,188 @@ def _table_rows(
     return rows
 
 
+def _format_reward(value: object) -> str:
```
joyliu-q: Why is value an object? What are your formatting here?

modal_training_gym/cli/run.py
```diff
@@ -98,6 +99,188 @@ def _table_rows(
...
+        return RunSummary.model_validate(payload)
+    except ValidationError as exc:
+        raise CLIError(
+            "Dashboard returned an invalid run summary.",
+            error="invalid_dashboard_response",
+            exit_code=ExitCode.BACKEND,
+        ) from exc
+
+
+def _validate_run_summaries(payload: object) -> list[RunSummary]:
```
joyliu-q: Isn't this function just a helper function for list comprehension? This is AI generated and not needed

tests/test_dashboard_runs_route.py
```diff
@@ -66,6 +66,33 @@ def test_runs_route_returns_typed_joined_summaries(fake_volume, monkeypatch, tmp
     assert summary["train_result"]["checkpoint_dir"] == ("/checkpoints/run-route-1")
 
 
+def test_get_run_route_returns_one_typed_summary(fake_volume, monkeypatch, tmp_path):
```
joyliu-q, line 72: When would this not return one typed summary? What are the tests trying to accomplish

modal_training_gym/cli/run.py
rhea24: it's because the type of rollout was dict[str, object] and we were passing in `rollout.get("mean")` in one of the callers. but actually i'll add a new pydantic type for rollout specifically

tests/test_dashboard_runs_route.py
rhea24: because it's a new endpoint just wanted to make sure there was coverage that it directly fetches the requested run instead of scanning the summary stores (since that was the main reason for adding the endpoint)

joyliu-q:
I think we can give the human UI dashboard layout more sauce to make it not a simple table

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -68,6 +91,33 @@
...
+        (runStatus && runStatus !== "running")
+      ) return;
+      void loadRun(id, controller.signal);
+    }, 5000);
+
+    return () => {
+      controller.abort();
+      window.clearInterval(interval);
+    };
+  });
```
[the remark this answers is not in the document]
rhea24: this is not how the order of operations in the backend is executed

modal_training_gym/cli/run.py
```diff
@@ -150,6 +397,28 @@ def run_group() -> None:
     """Inspect and manage training runs."""
 
 
+@run_group.command(
+    "get",
+    help="Show status and top-level metadata for a single run.",
+    epilog=(
+        "Examples:\n"
+        "  training-gym run get run_8f2a\n"
```
bokeefe123: Probably better to have the name here match ones a user would expect to see something like `brave-falcon-3fa8` or similar

bokeefe123 (approved):
small nit but otherwise lgtm

modal_training_gym/_dashboard.py
```diff
@@ -736,6 +740,25 @@ async def runs(
...
+                detail=f"Training run {training_run_id!r} not found",
+            )
+        try:
+            result = await vol_get(
+                MetadataStore.TRAIN_RESULTS,
+                training_run_id,
+                is_async=True,
+            )
+        except KeyError:
+            result = None
```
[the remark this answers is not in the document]
rhea24: tighter catches are better

PR #297, opened 2026-07-30
Skip GLM 4.7 in CI runs

anli5005:
Skips validating GLM 4.7 during CI runs in an effort to cut CI time, achieved by moving CI to reading from an explicit model catalog instead of generating it at runtime.


joyliu-q (approved):
I like this allow list approach

PR #298, opened 2026-07-30
Add skip-in-ci which will skip checks for larger more expensive models (should be ran more infrequently)

joyliu-q:


anli5005:
Closing as we decided to go with the approach in #297 after modifications to that PR

PR #302, opened 2026-07-31
Report model-validation step time instead of wall clock in comment

joyliu-q:
Model-validation reported wall-clock `training_run.duration_seconds`, which includes queue time plus model download/conversion — variable with compute availability rather than gym performance (e.g. 87m40s wall clock vs 6m10s of step time).

`total_duration_s` keeps its meaning (wall clock); the step-time sum is derived at comment-generation time:

```python
def _total_step_time_s(result) -> float:
    return sum(step["duration_s"] or 0 for step in (result.step_times or {}).values())
```

- Results table "Duration" column → "Step time", using the sum.
- Per-model details table: trailing row is now "Total step time", and it's omitted when there's only one step (it would just repeat the single step row).
- `print_summary` prints both step time and wall-clock duration.

## Checklist

N/A — script change, not an example.

Link to Devin session: https://modal.devinenterprise.com/sessions/b3d493270ae64bfd898ac264143d257e
Requested by: @joyliu-q

### DIFF
--- a/scripts/validate_model_configs.py
+++ b/scripts/validate_model_configs.py
@@ -54,16 +54,15 @@ def _substep_label(name: str) -> str:
     return _SUBSTEP_LABELS.get(name, name.replace("_", " "))
 
 
-def _sum_step_durations(
-    step_times: dict[str, dict[str, int | None]] | None,
-) -> float:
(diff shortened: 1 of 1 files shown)


anli5005 (approved):
LGTM

PR #305, opened 2026-08-01
dashboard: opt-in Modal proxy auth; authenticate ingestion before run lookup

pedramr:
### 1. Ingestion routes authenticate before the run lookup

The three ingestion POSTs (`/api/framework-status`, `/api/training-rollouts`, `/api/advantage-distributions`) looked up the run before checking the per-run bearer token. An unknown run returned 404 and a bad token returned 403, so anyone could tell valid training-run ids apart without credentials.

The token check now runs first. A run with no token on record still performs a `compare_digest` against a dummy value instead of returning early, so all rejections look the same.

### 2. `TRAINING_GYM_DASHBOARD_REQUIRES_PROXY_AUTH`

Set this to `1` in the shell that runs `training-gym setup` (or `modal deploy dashboards/app.py`) to deploy the dashboard with `requires_proxy_auth=True`. That is the same edge protection the serving path already offers through `DeploymentConfig(unauthenticated=False)`. It defaults to off.

The value is read at decorator time in the deploying process, so a Secret attached to the remote function has no effect on it. Only `1`, `true`, `0`, `false`, and unset are accepted; anything else raises, since a typo that quietly meant "off" would deploy an open dashboard. The choice is sticky: an explicit value persists to `~/.training-gym.toml`, and unset falls back to the persisted choice, so redeploying from a shell that doesn't export the variable can't silently reopen a locked dashboard — downgrading takes an explicit `0`. `training-gym setup` prints the edge-auth state both ways, so a machine without the persisted choice deploying open is never silent.

Browsers can't send proxy-auth headers, so this works alongside `training-gym set-password` rather than replacing it.

### 3. Dashboard clients can pass Modal edge auth

A new `common/proxy_auth.py`, extracted from `deployment.py`, adds the workspace `wk-`/`ws-` pair from `MODAL_KEY`/`MODAL_SECRET` (environment, or `~/.training-gym.toml` locally). It sends both headers or neither, never one — and only to `https` URLs, on Modal-served hostnames (`*.modal.run`, `*.modal.direct`) or hosts named in `TRAINING_GYM_PROXY_AUTH_HOSTS`: the pair is a workspace credential and the dashboard URL is operator-supplied configuration, so anywhere else gets the request without it. Naming a host says where the pair may go, not that it may go in the clear — `https` is required there too, loopback excepted since there is no network to observe. Three callers use it:

- Both status reporters, `common/status_reporter.py` and `frameworks/slime/reporting.py`. Only patching the first would have quietly dropped slime's rollouts and advantage distributions. Both now refuse redirects rather than replaying the run token and the pair at whatever origin a `Location` header names, and both warn on stderr (rate-limited) for a 401, a 403, or a refused redirect, so a misconfigured dashboard doesn't silently stop receiving updates. A 401 whose pair was withheld by the origin check says so, instead of telling the operator to set what is already set.
- The CLI `DashboardClient`, through a request event hook tied to the configured origin. This needs to be a hook rather than an `httpx.Auth` flow because it has to run on every redirect hop — httpx drops `Authorization` when the origin changes but carries custom headers over, so the pair needs its own check.
- The slime and miles `download` and `convert_checkpoint` functions, which report phase status but never mounted `proxy_auth_secrets()`, so those phases stopped reporting whenever the flag was on.

`deployment.py`, where the helper came from, keeps using it for serve/eval calls; its 401 hint now distinguishes a configured pair withheld from a non-Modal serve URL from no pair at all, naming `TRAINING_GYM_PROXY_AUTH_HOSTS` for the former instead of advising the operator to set what is already set.

One deliberate break to note in the release notes: `_modal_proxy_auth_headers` survives as an alias, but it now takes a URL, so a zero-argument call raises `TypeError` — which matters because the tutorials import it by name inside reward/generate functions that cloudpickle ships to rollout workers, so user code copied from an earlier tutorial fails there rather than at import. The in-repo tutorial call sites are updated; a compatible zero-arg signature was rejected because it would have to either send the pair unconditionally (the leak this guard exists to close) or return nothing and turn the break into a silent 401.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/305" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/_dashboard.py
+++ b/modal_training_gym/_dashboard.py
@@ -128,6 +128,53 @@ PASSWORD_EXEMPT_PATHS = frozenset(
     }
 )
 
+# Opt-in Modal edge auth: with this set, Modal rejects any request without a
+# workspace proxy-auth pair before app code runs. The reporters and the CLI
+# send it from MODAL_KEY/MODAL_SECRET (see common/proxy_auth.py); launchers
+# forward it into train containers via proxy_auth_secrets().
+REQUIRES_PROXY_AUTH_ENV = "TRAINING_GYM_DASHBOARD_REQUIRES_PROXY_AUTH"
+
+# Only ever the *expected* side of a comparison, so publishing it is safe.
+_MISSING_TOKEN_DUMMY = "training-gym-missing-token-dummy-never-issued"
+
+
+def _requires_proxy_auth() -> bool:
+    """Resolve the opt-in edge-auth flag, stickily.
+
+    Read in the *deploying* process at decorator time. Anything but
+    1/true/0/false/unset raises, because a typo that read as "off" would
+    deploy an open dashboard. An explicit value is persisted to
+    ``~/.training-gym.toml`` and unset falls back to the persisted choice,
+    so a redeploy from a shell that doesn't export the variable cannot
+    silently reopen the dashboard — downgrading takes an explicit ``0``.
+    Containers only parse the env var and never touch the config file.
+    """
+    raw = os.environ.get(REQUIRES_PROXY_AUTH_ENV, "").strip().lower()
+    if raw in ("1", "true"):
+        explicit = True
+    elif raw in ("0", "false"):
+        explicit = False
+    elif raw == "":
+        explicit = None
+    else:
+        raise ValueError(
+            f"{REQUIRES_PROXY_AUTH_ENV} must be 1/true or 0/false/unset, got {raw!r}"
+        )
+
+    if not _is_local():
+        return bool(explicit)
+
+    from modal_training_gym.common.config import (
+        get_dashboard_requires_proxy_auth,
+        save_dashboard_requires_proxy_auth,
+    )
+
+    if explicit is None:
+        return get_dashboard_requires_proxy_auth()
+    if explicit != get_dashboard_requires_proxy_auth():
+        save_dashboard_requires_proxy_auth(explicit)
+    return explicit
+
 
 def _is_local() -> bool:
     """True when we're not running inside a Modal container."""
@@ -355,7 +402,7 @@ def reconcile() -> None:
     min_containers=1,
     secrets=_function_secrets(),
 )
-@modal.asgi_app()
+@modal.asgi_app(requires_proxy_auth=_requires_proxy_auth())
 def fastapi_app():
     import base64
     import binascii
@@ -684,6 +731,12 @@ def fastapi_app():
     async def _require_framework_status_token(
         training_run_id: str, authorization: str | None
     ) -> None:
+        """403 unless ``authorization`` carries the run's status token.
+
+        Handlers must call this *before* any lookup that can 404, or the
+        status code tells an anonymous caller which run ids exist. For the
+        same reason an unknown run is indistinguishable from a wrong token.
+        """
         try:
             expected_token = str(
                 (
@@ -696,9 +749,13 @@ def fastapi_app():
             )
         except KeyError:
             expected_token = ""
-        if not expected_token or not _secrets.compare_digest(
-            _bearer_token(authorization), expected_token
-        ):
+        supplied = _bearer_token(authorization)
+        if not expected_token:
+            # Spend the same comparison an existing token would, then refuse
+            # regardless of its outcome.
+            _secrets.compare_digest(supplied, _MISSING_TOKEN_DUMMY)
+            raise HTTPException(status_code=403, detail="Invalid status token")
+        if not _secrets.compare_digest(supplied, expected_token):
             raise HTTPException(status_code=403, detail="Invalid status token")
 
     async def _get_run_or_404(training_run_id: str) -> TrainingRun:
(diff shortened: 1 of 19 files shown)


modal_training_gym/_dashboard.py
```diff
@@ -128,6 +128,53 @@ def _build_image() -> modal.Image:
...
+    """Resolve the opt-in edge-auth flag, stickily.
+
+    Read in the *deploying* process at decorator time. Anything but
+    1/true/0/false/unset raises, because a typo that read as "off" would
+    deploy an open dashboard. An explicit value is persisted to
+    ``~/.training-gym.toml`` and unset falls back to the persisted choice,
+    so a redeploy from a shell that doesn't export the variable cannot
+    silently reopen the dashboard — downgrading takes an explicit ``0``.
+    Containers only parse the env var and never touch the config file.
+    """
```
anli5005, line 151: It might be more ergonomic to expose this setting as a command-line flag on `training-gym setup` instead of an environment variable, perhaps something along these lines:

* `training-gym setup --proxy-auth` deploys the dashboard and configures it to require proxy auth.
* `training-gym setup --no-proxy-auth` deploys the dashboard and disables proxy auth if it was enabled.
* `training-gym setup` without any flags first checks if the dashboard was last deployed with proxy auth. If it was, it fails and tells the user to pass in either `--proxy-auth` or `--no-proxy-auth`; otherwise, it defaults to deploying without proxy auth.

An environment variable would traditionally be required for running `modal deploy` directly, but since this path is not supported this should be fine.

modal_training_gym/common/proxy_auth.py
```diff
@@ -0,0 +1,175 @@
...
+    except ValueError:
+        return False
+    host = (parts.hostname or "").lower()
+    if not host:
+        return False
+    if parts.scheme != "https" and host not in _LOOPBACK_HOSTS:
+        return False
+    if host in _extra_hosts():
+        return True
+    return host.endswith(_MODAL_HOST_SUFFIXES)
```
anli5005, line 59: This check works for most use cases but I worry if it's over-optimized. I can think of two main scenarios where you'd end up sending a proxy token to a non-Modal URL:

1. You have some sort of custom proxy or domain for the training gym dashboard, and you've modified the training gym URL in your local config.
2. You call `_modal_proxy_auth_headers` from your reward function and explicitly send it off to some other domain.

In both cases the intent is clear enough that I'm not sure the additional developer friction of needing to set `TRAINING_GYM_PROXY_AUTH_HOSTS` (and potentially override `_LOOPBACK_HOSTS`) is additive here. I would suggest removing this check and the associated logic to simplify the token logic and allow us to revert `_modal_proxy_auth_headers` to its original call signature.

modal_training_gym/common/proxy_auth.py
```diff
@@ -0,0 +1,175 @@
...
+    return {}
+
+
+class _RefuseRedirects(HTTPRedirectHandler):
+    """urllib replays the original headers — run token and proxy pair — at
+    whatever origin a redirect names. The ingestion endpoints never redirect,
+    so treat a 3xx as the misconfiguration it is."""
+
+    def redirect_request(self, req, fp, code, msg, headers, newurl):
+        return None
```
anli5005, line 93: Similarly I'm not sure if the `RefuseRedirects` code is additive here, as this scenario will never occur in practice provided one doesn't change the dashboard URL manually. It might be best to remove this logic as well to further reduce over-optimization.

modal_training_gym/_dashboard.py
```diff
@@ -696,9 +749,13 @@ async def _require_framework_status_token(
...
-        if not expected_token or not _secrets.compare_digest(
-            _bearer_token(authorization), expected_token
-        ):
+        supplied = _bearer_token(authorization)
+        if not expected_token:
+            # Spend the same comparison an existing token would, then refuse
+            # regardless of its outcome.
+            _secrets.compare_digest(supplied, _MISSING_TOKEN_DUMMY)
+            raise HTTPException(status_code=403, detail="Invalid status token")
+        if not _secrets.compare_digest(supplied, expected_token):
```
anli5005, line 758: I like the way you implemented this check; if you're able to split the run lookup changes into a separate PR we can approve it and merge it in!

anli5005 (commented):
Hey @pedramr - thanks for your contribution! We've been working to improve the auth story on the Training Gym, so a PR like this is a huge help.

We have a few comments about some of the implementation details if you'd like to take a look; otherwise, we're happy to use this PR as a base and hack on things on our end.

The changes to our run lookup endpoint in particular look good to go; if you're able to separate those out into another PR we're happy to merge it in!

modal_training_gym/_dashboard.py
pedramr: Thanks for the review! Moved this part to https://github.com/modal-projects/training-gym/pull/310. And I agree with the suggested changes (e.g., having a new `training-gym setup --proxy-auth`),  just wasn't quite sure how much to make this a first-party feature, and you have a much better vision for it and this was more of a placeholder PR, so feel free to close this and add it in a more cohesive way.

PR #307, opened 2026-08-03
Default rollout_health_check_first_wait to 300s

ayushr2:
slime's rollout health monitor defaults its first-probe grace period (`--rollout-health-check-first-wait`) to 0, so it probes `/health_generate` during rollout-0's cold model compilation. The busy engine returns 503, the monitor kills it on the first failure, and with `--use-fault-tolerance` it is not relaunched — failing the run. The compile is slower under gVisor, where a first_wait of 0 reliably triggers this.

This adds a `rollout_health_check_first_wait` field to `SlimeRecipe` defaulting to 300s, which covers the cold-compile window. slime's own help recommends raising it "significantly when using deepgemm". Under runc it only delays the first health probe.

Overridable per-recipe if a model needs a different value.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/307" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/train_recipes/slime_recipe/recipe.py
+++ b/modal_training_gym/train_recipes/slime_recipe/recipe.py
@@ -237,6 +237,21 @@ class SlimeRecipe(BaseTrainRecipe):
     rollout_stop_token_ids : list[int] | None
         Extra token ids that terminate generation.
 
+    ## Fault Tolerance and Health Checks
+    use_fault_tolerance : bool
+        Enable slime's fault tolerance so the run can recover from worker
+        failures.
+    rollout_health_check_interval : int
+        Interval in seconds between rollout engine /health_generate checks
+        during generate/eval.
+    rollout_health_check_timeout : int
+        Timeout in seconds to wait for a rollout engine /health_generate
+        response before killing it.
+    rollout_health_check_first_wait : int
+        Initial grace period (in seconds) before starting health checks. This
+        allows time for model compilation and initialization. Increase this
+        value significantly when using deepgemm.
+
     ## Checkpointing
 
     save : str
@@ -251,9 +266,6 @@ class SlimeRecipe(BaseTrainRecipe):
     megatron_to_hf_mode : str
         Mode used to export saved Megatron checkpoints back to HF format;
         empty disables the export step.
-    use_fault_tolerance : bool
-        Enable slime's fault tolerance so the run can recover from worker
-        failures.
     freeze_params_name_list : list[str] | None
         Regex patterns (matched with ``re.search``) of parameter names to
         freeze, e.g. a VL model's vision tower so RL only updates the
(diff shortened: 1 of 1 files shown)


modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -236,6 +236,10 @@ class SlimeRecipe(BaseTrainRecipe):
         Nucleus-sampling top-p for rollout generation.
     rollout_stop_token_ids : list[int] | None
         Extra token ids that terminate generation.
+    rollout_health_check_first_wait : int
+        Seconds before the rollout health monitor's first ``/health_generate``
+        probe; above slime's default of 0 so cold model compilation finishes
```
anli5005: It looks like [the default is already 300](https://github.com/THUDM/slime/blob/main/docs/en/advanced/fault-tolerance.md#:~:text=Default%3A%20300%20seconds.)

modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -236,6 +236,10 @@ class SlimeRecipe(BaseTrainRecipe):
         Nucleus-sampling top-p for rollout generation.
     rollout_stop_token_ids : list[int] | None
         Extra token ids that terminate generation.
+    rollout_health_check_first_wait : int
```
anli5005, line 250: nit: Maybe this would fit better under a different section, like Cluster and Parallelism? Not sure

modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -534,6 +538,7 @@ class SlimeRecipe(BaseTrainRecipe):
     rollout_top_p: float = 1.0
     rollout_stop_token_ids: list[int] | None = None
     sglang_mem_fraction_static: float = 0.75
+    rollout_health_check_first_wait: int = 300
```
anli5005, line 554: Not a blocker but maybe should we add the other [health check flags while we're at it?](https://github.com/THUDM/slime/blob/main/docs/en/advanced/fault-tolerance.md#rollout-health-checks)

modal_training_gym/train_recipes/slime_recipe/recipe.py
ayushr2: Somehow the code defaults it to 0... https://github.com/THUDM/slime/blob/66034ab772ab9e4e181ee24a26783681a6ce6b68/slime/utils/arguments.py#L582

modal_training_gym/train_recipes/slime_recipe/recipe.py
ayushr2: Hmm all the rollout specific flags seem to be in this section... Do you have a strong preference as to which one?

ayushr2: Created a new section for this, done.

modal_training_gym/train_recipes/slime_recipe/recipe.py
ayushr2: Done.

modal_training_gym/train_recipes/slime_recipe/recipe.py
ayushr2: Hmm pretty sure the defaults in the doc page you listed is stale/incorrect because the code is doing something else. The Slime default has bit me during testing...

anli5005 (approved):
LGTM

PR #310, opened 2026-08-03
dashboard: authenticate ingestion before run lookup

pedramr:
The three ingestion POSTs (`/api/framework-status`, `/api/training-rollouts`, `/api/advantage-distributions`) looked up the run before checking the per-run bearer token. An unknown run returned 404 and a bad token returned 403, so anyone could tell valid training-run ids apart without credentials.

The token check now runs first. A run with no token on record still performs a `compare_digest` against a dummy expected value instead of returning early, so all rejections look the same — same 403, same body, same comparison cost. The dummy only ever sits on the expected side of the comparison, so presenting it as a bearer token still fails.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/310" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/_dashboard.py
+++ b/modal_training_gym/_dashboard.py
@@ -128,6 +128,9 @@ PASSWORD_EXEMPT_PATHS = frozenset(
     }
 )
 
+# Only ever the *expected* side of a comparison, so publishing it is safe.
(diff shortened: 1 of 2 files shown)


anli5005 (approved):
Looks good to me, thanks for the contribution!

PR #311, opened 2026-08-03
Support run kill cli

rhea24:
Support `run kill` CLI command

Screenshot 2026-08-05 at 12 30 28 PM

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/311" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


modal_training_gym/cli/run.py
```diff
@@ -430,14 +449,211 @@ def download_run_traces(
...
+    runs_to_kill = [
+        run_report for run_report in run_reports if run_report["action"] == "would_kill"
+    ]
+    error_count = sum(
+        (
+            run_report["skip_reason"] is not None
+            and run_report["skip_reason"] != "already_terminal"
+        )
+        for run_report in run_reports
+    )
```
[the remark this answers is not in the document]
rhea24: the cli command is meant to kill the underlying modal app so this is intentional - if underlying modal app is already killed then it shouldnt do anything

modal_training_gym/common/run.py
```diff
@@ -278,6 +278,34 @@ async def _run() -> TrainingRun:
...
+    status: TrainingRunStatus,
+    reason: str,
+    ended_at: int | None = None,
+) -> TrainingRunStatus:
+    """Persist a terminal status if the canonical run is still running."""
+    run = TrainingRun.from_id(run_id)
+    if not isinstance(run, TrainingRun):
+        raise TypeError("TrainingRun.from_id returned an asynchronous result")
+    if run.status != TrainingRunStatus.RUNNING:
+        return run.status
```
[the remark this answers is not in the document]
rhea24: pending is not a possible state for `TrainingRunStatus`

modal_training_gym/cli/run.py
```diff
@@ -48,6 +50,19 @@
...
+    run_id: str
+    modal_app_id: str | None
+    modal_app_live: bool | None
+    status: str
+    current_step: int | None
+    total_steps: int | None
+    step_unit: str
+    duration_seconds: int | None
+    action: Literal["would_kill", "skipped", "killed"]
+    skip_reason: str | None
```
joyliu-q, line 63: What is skip reason?

modal_training_gym/cli/run.py
```diff
@@ -434,14 +449,198 @@ def download_run_traces(
...
+                        not_found_error=CLIError(
+                            f"Training run {run_id!r} was not found.",
+                            error="run_not_found",
+                            exit_code=ExitCode.NOT_FOUND,
+                            run_id=run_id,
+                            hint="training-gym run list",
+                        ),
+                    )
+                )
+            )
```
joyliu-q, line 552: I dont think kill should mass delete training runs, this seems dangerous

modal_training_gym/_dashboard.py
```diff
@@ -733,6 +752,67 @@ async def get_run(training_run_id: str):
...
+        training_run_id: str,
+        authorization: str | None = Header(default=None),
+    ):
+        _require_mutation_auth(authorization)
+        run = await _get_run_or_404(training_run_id)
+        if run.status != TrainingRunStatus.RUNNING:
+            return JSONResponse(
+                {
+                    "action": "skipped",
+                    "skip_reason": "already_terminal",
```
joyliu-q, line 766: This doesn't really make sense as a "skip" reason

modal_training_gym/_dashboard.py
```diff
@@ -733,6 +752,67 @@ async def get_run(training_run_id: str):
...
+                    "status": run.status.value,
+                }
+            )
+        if not run.modal_app_id:
+            return JSONResponse(
+                {
+                    "action": "skipped",
+                    "skip_reason": "missing_modal_app_id",
```
joyliu-q, line 774: Seems like an error, not a skip reason

modal_training_gym/cli/run.py
rhea24: oh i thought that's what we wanted it to do? what do you think instead?

modal_training_gym/cli/run.py
rhea24: there are cases when a run that user wanted killed cant actually be killed because they're already stopped and i didnt think it should error out completely - i thought best impl would be one where we kill all the ones the user specified and if not for any of them, we print out why we didnt kill it (`skip_reason`)so the user knows it was already killed

modal_training_gym/_dashboard.py
rhea24: it doesnt? i thought it would because it means you're trying to kill a run that's already ended

modal_training_gym/_dashboard.py
rhea24: but this could just mean the modal app for the run you specified was already terminated right?

modal_training_gym/_dashboard.py
```diff
@@ -418,6 +424,19 @@ def _password_ok(authorization: str | None) -> bool:
...
+                status_code=403,
+                detail="Set a dashboard password before using mutation endpoints",
+            )
+        if not _password_ok(authorization):
+            raise HTTPException(
+                status_code=401,
+                detail="Dashboard authentication was rejected",
+                headers={"WWW-Authenticate": 'Basic realm="training-gym"'},
+            )
+
```
joyliu-q, line 439: This requires `DASHBOARD_PASSWORD` for the kill endpoint, so proxy-auth-only dashboards can't use run kill even though Modal already authenticated the request at the proxy layer.

It also leaks what config state one has with the 403 detail "Set a dashboard password before using mutation endpoints".

Maybe this is a sign that `kill` might not be a good command for those cases.

Thoughts on pausing this PR for now, and having `training-gym run list` also show the modal app id so the agent can just kill it if it has the correct modal credentials?

rhea24: oo yeah i agree, sg - that way the agent can still kill if they want but not have to deal with this extra overhead

PR #313, opened 2026-08-04
Add option to require proxy auth for gym dashboard

anli5005:
This PR introduces an opt-in `--proxy-auth` flag on `training-gym setup`, which allows users to deploy the dashboard with proxy auth enabled, in a manner similar to #305. Subsequent deploys will require either passing in `--proxy-auth` or `--no-proxy-auth`.

The dashboard client has been extended to handle passing in proxy auth using the existing config setup.


modal_training_gym/cli/setup.py
```diff
@@ -24,18 +24,46 @@
...
+        return proxy_auth
+
+    from modal_training_gym.common.config import get_dashboard_proxy_auth
+
+    if get_dashboard_proxy_auth() is True:
+        raise ProxyAuthChoiceRequired(
+            "The dashboard was last deployed with proxy auth. "
+            "Pass either --proxy-auth or --no-proxy-auth explicitly."
+        )
+    return False
```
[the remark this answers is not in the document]
anli5005: This is acceptable behavior; we want to require users to make an explicit decision by running the command themselves, and it's unlikely they would end up in a situation where the dashboard is not deployed yet `[dashboard].proxy_auth = true`.

modal_training_gym/cli/builtin.py
```diff
@@ -8,11 +8,30 @@
...
+@click.option(
+    "--proxy-auth",
+    is_flag=True,
+    help="Require Modal proxy authentication for the dashboard.",
+)
+@click.option(
+    "--no-proxy-auth",
+    is_flag=True,
+    help="Deploy the dashboard without Modal proxy authentication.",
+)
```
joyliu-q, line 20: Why is there both `--proxy-auth` and `--no-proxy-auth`

joyliu-q: Shouldnt no proxy auth be already assumed as a default w/o an extra flag

anli5005: If the user previously deployed with `--proxy-auth`, we require them to pass in `--no-proxy-auth` to make their intent clear

modal_training_gym/common/status_reporter.py
```diff
@@ -89,7 +89,14 @@ def _post(item: dict[str, Any]) -> None:
...
     body = json.dumps(item, default=str).encode("utf-8")
-    headers = {"Content-Type": "application/json"}
+
+    from modal_training_gym.common.config import modal_proxy_auth_headers
+
+    headers = {
+        "Content-Type": "application/json",
+        **modal_proxy_auth_headers(),
+    }
+
```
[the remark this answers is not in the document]
joyliu-q: This seems valid, we should fix outside of this PR

joyliu-q (approved):

PR #191, opened 2026-06-26
Add slime training step timings

cskitty:
Add step time logging for end-to-end wall-clock step lifecycle  

Timestamps are emitted from patched slime loop: start (before rollout generation) and finish (before "if should_run_periodic_action(rollout_id, args.save_interval, num_rollout_per_epoch, args.num_rollout):") are written onto new Modal Dict (for all runs). At the end of every training run, start/finish/duration for all steps are saved to TrainingRun.step_times and flushed from Modal Dict.

PR also adds wandb support to validator runs.

Tested on two rollouts of Qwen3-0.6B and Qwen3-4B.

Example TrainingRun.step_times:
`{
    "1": {"start": 1782513857, "end": 1782513900, "duration_s": 43},
    "2": {"start": 1782513902, "end": 1782513928, "duration_s": 26},
}`

Included in my step_times:

sleep_time
+ update_weights_time
+ wake_up_time
+ data_preprocess_time
+ train_wait_time
+ train_time

Only train_wait_time + train_time are included in [wandb perf/step_time](https://github.com/THUDM/slime/blob/d719f036/slime/utils/train_metric_utils.py#L44-L48).

### Qwen 3 0.6B

TrainingRun:

https://modal.com/api/volumes/modal-labs/melody-dev/training-gym-metadata/files/content?path=training-runs%2Fslow-core-50daaf0de2a5.json

W&B:

https://api.wandb.ai/links/cskitty-mit/2qlesc33

"step_times": {
    "1": {"start": 1782513857, "end": 1782513900, "duration_s": 43},
    "2": {"start": 1782513902, "end": 1782513928, "duration_s": 26},
}

STEP 2:

STEP_TIME:
sleep_time:           2.056785
update_weights_time:  0.540130
wake_up_time:         0.969507
data_preprocess_time: 0.004356
train_wait_time:     17.958546
train_time:           4.317374
2.056785 + 0.540130 + 0.969507 + 0.004356 + 17.958546 + 4.317374
= 25.846698s
≈ 26s

WANDB:
train_wait_time + train_time
= 17.958546 + 4.317374
= 22.275920s

"step_times": {
  "1": {"start": 1782522497, "end": 1782522538, "duration_s": 41},
  "2": {"start": 1782522541, "end": 1782522558, "duration_s": 17}
}

01:09:01  step_times["2"].start recorded by dashboard

01:09:07  rollout perf 1 logged by RolloutManager
01:09:08  Timer wake_up start
01:09:09  Timer wake_up end
[...]

PR #352, opened 2026-08-07
Add Miles CI & refactor CI

zhouhelena1:
adds miles validation ci (consolidated with slime ci) but doesn't run on anything today

tested by temporarily putting kimi in `MILES_VALIDATABLE_MODELS`, it shows up in validate step, then i cancelled it.

original pr/comments were here, but ci was queuing forever so I closed and reopened pr: https://github.com/modal-projects/training-gym/pull/337

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/352" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


.github/workflows/validate-models.yml
```diff
@@ -1,24 +1,39 @@
...
 
 permissions:
   contents: read
 
 concurrency:
   group: validate-models-${{ github.ref }}
-  cancel-in-progress: true
+  # Superseded PR runs are free to cancel, but a dispatched run may be a
+  # 16 x 8 H200 job, so a second dispatch must never abort one already training.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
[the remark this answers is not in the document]
zhouhelena1: Good catch — fixed in 65f134e7 by appending `github.run_id` to the group for non-PR events, so each manual dispatch gets its own group and runs in parallel. PR runs keep the shared per-ref group and supersede-and-cancel behavior.

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
-    ("Qwen3-ASR-1.7B", Qwen3_ASR_1_7B),
-    ("Qwen3-VL-8B-Instruct", Qwen3_VL_8B),
-    ("Qwen3.6-35B-A3B", Qwen3_6_35B),
+
+class ValidationFramework(StrEnum):
+    """Framework a validation run trains on."""
+
+    SLIME = "slime"
+    MILES = "miles"
+
```
joyliu-q: Don't make a new enum, just directly use 

```
class Framework(str, Enum):
    # ``str`` mixin so a Framework serializes straight to its value in plain
    # ``json.dumps`` (e.g. TrainResult, whose ``asdict`` payload doesn't coerce
    # enums) — matching the SlimeStatus/MilesStatus convention.
    SLIME = "slime"
    MILES = "miles"
```

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+
+@dataclass(frozen=True)
+class ValidationTarget:
+    """One model/framework pair the validation harness knows how to run."""
+
+    name: str
+    """Short name used as ``check --model`` and as the CI matrix entry."""
```
joyliu-q: Why are the comments after the field?

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+
+    model_config: type[ModelConfig]
+
+    framework: ValidationFramework
+    """Which framework's ``get_base_recipe`` trains this model."""
```
joyliu-q: Use # not """

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+
+    @property
+    def model_name(self) -> str:
+        """The HF repo id, e.g. ``moonshotai/Kimi-K2.5``."""
+        return self.model_config.model_name
+
+    def matches(self, name: str) -> bool:
+        """Whether ``name`` refers to this target, by short name or repo id."""
+        wanted = name.strip().lower()
+        return wanted in (self.name.lower(), self.model_name.lower())
```
joyliu-q: This feels like overengineering, in most cases the field is automatically passed in and don't need to be matched against multiple possible values

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+    name: str
+    """Short name used as ``check --model`` and as the CI matrix entry."""
+
+    model_config: type[ModelConfig]
+
+    framework: ValidationFramework
+    """Which framework's ``get_base_recipe`` trains this model."""
+
+    gates_prs: bool = True
+    """Whether a pull request may launch this run.
```
joyliu-q: Rename to `ci_enabled`

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+
+class ValidationFramework(StrEnum):
+    """Framework a validation run trains on."""
+
+    SLIME = "slime"
+    MILES = "miles"
+
+
+@dataclass(frozen=True)
+class ValidationTarget:
```
joyliu-q: `_ValidationConfig` use underscore since it is only used internally and not in docs.

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+        """The HF repo id, e.g. ``moonshotai/Kimi-K2.5``."""
+        return self.model_config.model_name
+
+    def matches(self, name: str) -> bool:
+        """Whether ``name`` refers to this target, by short name or repo id."""
+        wanted = name.strip().lower()
+        return wanted in (self.name.lower(), self.model_name.lower())
+
+
+VALIDATION_TARGETS: tuple[ValidationTarget, ...] = (
```
joyliu-q: make it a `set`

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+    *,
+    gating_only: bool = False,
+) -> tuple[ValidationTarget, ...]:
+    """Registry entries, optionally narrowed to one framework or the PR set."""
+    return tuple(
+        target
+        for target in VALIDATION_TARGETS
+        if (framework is None or target.framework == framework)
+        and (not gating_only or target.gates_prs)
+    )
```
joyliu-q: Why are we tupling

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+    ValidationTarget(
+        "Kimi-K2.6", Kimi_K2_6, ValidationFramework.MILES, gates_prs=False
+    ),
 )
+
+
+def validation_targets(
+    framework: ValidationFramework | None = None,
+    *,
+    gating_only: bool = False,
```
joyliu-q: When would one pass in `gating_only` to True?

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+    )
+
+
+def find_validation_target(name: str) -> ValidationTarget:
+    """Look up a target by short name or HF repo id, case-insensitively."""
+    for target in VALIDATION_TARGETS:
+        if target.matches(name):
+            return target
+    available = ", ".join(target.name for target in VALIDATION_TARGETS)
+    raise ValueError(f"unknown model {name!r}; available: {available}")
```
joyliu-q: This needs to be a helper class fn on `_ValidationConfig`/`ValidationTarget`

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +30,112 @@
...
+    """
+    seen: dict[str, ValidationTarget] = {}
+    for target in targets:
+        for key in (target.name.lower(), target.model_name.lower()):
+            other = seen.get(key)
+            if other is not None and other is not target:
+                raise ValueError(
+                    f"validation targets {other.name!r} ({other.model_name}) and "
+                    f"{target.name!r} ({target.model_name}) both answer to {key!r}"
+                )
```
joyliu-q: Seems like this should be added to testing, not library code

This assertion also runs on every `import` which is pretty bad

scripts/validation_backends/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+    non_colocated: bool = False
+
+    def set_fields(self) -> tuple[str, ...]:
+        """Names of the overrides the caller actually passed."""
+        return tuple(
+            field.name
+            for field in fields(self)
+            if getattr(self, field.name) != field.default
+        )
+
```
joyliu-q: Why would we set recipe overrides? The default receipt out of the box should work on import

scripts/validation_backends/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _flag(field_name: str) -> str:
+    return "--" + field_name.replace("_", "-")
```
joyliu-q: Helper function but only used once. Please inline this function and self-review for this in the future.

scripts/validation_backends/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ValidationBackend(ABC):
```
joyliu-q: This is super overengineered.  This whole thing can just be a function, cased on slime or miles, instead of making an ABC.

scripts/validation_backends/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+
+        return MilesValidationBackend()
+    raise TrainingGymConfigError(f"no validation backend for framework {framework!r}")
+
+
+__all__ = [
+    "RecipeOverrides",
+    "ValidationBackend",
+    "backend_for",
+]
```
joyliu-q: Why

scripts/validation_backends/miles.py
```diff
@@ -0,0 +1,73 @@
...
+        overrides: RecipeOverrides,
+    ) -> MilesRecipe:
+        recipe = MilesRecipe.get_base_recipe(model_config)
+        if recipe is None:
+            raise TrainingGymConfigError(
+                f"no base miles recipe for model {model_config.model_name!r}; "
+                f"{target.name} is registered as a miles validation target"
+            )
+        if overrides.docker_image is not None:
+            recipe.docker_image = overrides.docker_image
```
joyliu-q: This is explicit and doesn't allow use to add other overrides which doesn't make sense

scripts/validation_backends/miles.py
```diff
@@ -0,0 +1,73 @@
...
+        step_count: int,
+    ) -> DatasetConfig:
+        """Validation dataset for miles models.
+
+        Every miles recipe today is a math-RL recipe scored by ``deepscaler``,
+        so they all validate against DAPO-Math-17k. Every rollout step consumes
+        ``rollout_batch_size`` prompts, so materialize enough for the whole run
+        rather than assuming the loader wraps epochs.
+        """
+        return DapoMath17kDataset(n_rows=recipe.rollout_batch_size * step_count)
```
joyliu-q: Not a needed function

scripts/validation_backends/miles.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def docker_image(self, recipe: MilesRecipe) -> str | None:
+        return recipe.docker_image
```
joyliu-q: Ideally we keep everything on the same miles docker image OR allow it to be specified in `MilesRecipe`.

If it is not possible to add Miles without bumping the image, we should bump the image to the latest working image first. I think we should only fragment the miles image used per recipe as a fail safe and if some recipes don't work.

scripts/validation_backends/slime.py
```diff
@@ -0,0 +1,151 @@
...
+        step_count: int,
+    ) -> DatasetConfig:
+        """Pick a validation dataset matching the base model's modality.
+
+        Audio models (Qwen3-ASR) need speech clips, so they get LibriSpeech;
+        everything else defaults to gsm8k.
+        """
+        if isinstance(model_config, Qwen3_ASR_1_7B):
+            return LibriSpeechASRDataset(n_rows=8)
+        return Gsm8kDataset(n_rows=10)
```
joyliu-q: Ditto here

scripts/validate_model_configs.py
```diff
@@ -140,170 +166,86 @@ def from_dict(cls, data: dict) -> "TutorialResult":
...
-        registry[model_config.model_name.lower()] = model_config
-    return registry
+def _override_field(flag: str) -> str:
+    """Map a ``check`` flag spelling ("docker-image") to a RecipeOverrides field."""
+    field_name = flag.strip().lstrip("-").replace("-", "_")
+    known = {field.name for field in fields(RecipeOverrides)}
+    if field_name not in known:
+        available = ", ".join(sorted(name.replace("_", "-") for name in known))
+        raise ValueError(f"unknown override {flag!r}; one of: {available}")
+    return field_name
```
joyliu-q: This is not needed here and overcomplicates user running validation because they need to think about what got overridden.

scripts/validate_model_configs.py
```diff
@@ -528,16 +479,16 @@ def summarize_results(
...
+    divider = "| --- | --- | --- | --- | --- | --- |"
+    empty = "| _no results_ | | | | | |"
     if baseline_dir is not None:
         header += " Delta |"
         divider += " --- |"
         empty += " |"
 
     lines = [
-        "<!-- validate-models-comment -->",
+        COMMENT_MARKER,
```
joyliu-q, line 478: Great driveby!

AGENTS.md
```diff
@@ -67,6 +70,16 @@ Every framework mounts three Modal Volumes:
...
+
+`common/models/validation.py` holds `VALIDATION_TARGETS`: each entry maps a model name to its `ModelConfig`, the framework whose `get_base_recipe` trains it, and `gates_prs`. The framework has to be declared — `SlimeRecipe.get_base_recipe` returns a recipe for any model it's asked about, so the recipe classes can't answer "is this model mine?".
+
+`scripts/validate_model_configs.py` owns everything framework-agnostic (CLI, result JSON, markdown summary, PR comment). `scripts/validation_backends/<framework>.py` owns the parts that differ: recipe lookup, validation dataset, and which `RecipeOverrides` the framework accepts — an override the backend doesn't list (`--docker-image` on slime, `--non-colocated` on miles) is an error, never a silent no-op. Adding a framework is one module here plus registry entries.
+
+`gates_prs=False` marks a model too expensive to launch from a PR (Kimi is 16 x 8 H200): still runnable by name from the CLI or `workflow_dispatch`, but `diff_impact.py` never puts it in a PR matrix. `tests/test_model_validation_registry.py` enforces that. `diff_impact.py` also scopes re-validation per framework, so a miles-only change doesn't re-run the slime set.
```
joyliu-q: update field name here

modal_training_gym/common/models/validation.py
zhouhelena1: to get the models that get run by PRs & gate merges, but I should flip this and have ci_only=True as the default

scripts/validation_backends/miles.py
zhouhelena1: got it, I'll remove docker image from this validation harness, so it's just declared in one place in MilesRecipe

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +21,71 @@
...
-    ("Qwen3-VL-8B-Instruct", Qwen3_VL_8B),
-    ("Qwen3.6-35B-A3B", Qwen3_6_35B),
-)
+
+class Framework(str, Enum):
+    # ``str`` mixin so a Framework serializes straight to its value in plain
+    # ``json.dumps`` (e.g. TrainResult, whose ``asdict`` payload doesn't coerce
+    # enums) — matching the SlimeStatus/MilesStatus convention.
+    SLIME = "slime"
+    MILES = "miles"
```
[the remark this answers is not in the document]
zhouhelena1: oops yes ill use that one

scripts/validation_backends/miles.py
```diff
@@ -0,0 +1,44 @@
...
+    the image a miles model trains on belongs in ``MilesRecipe``, not in the
+    validation harness.
+    """
+    recipe = MilesRecipe.get_base_recipe(model_config)
+    if recipe is None:
+        raise TrainingGymConfigError(
+            f"no base miles recipe for model {model_config.model_name!r}, "
+            "which is registered as a miles validation target"
+        )
+    return recipe, DapoMath17kDataset(n_rows=recipe.rollout_batch_size * step_count)
```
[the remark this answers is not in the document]
zhouhelena1: Checked: miles never had this override — `VALIDATION_EPHEMERAL_DISK_MIB` predates any miles work (#191, when the registry was slime-only), `MilesRecipe` has no `train_function_kwargs` field, and the miles launcher passes no `ephemeral_disk` to `train` or `convert_checkpoint`. It also wouldn't bite conversion: `hf_checkpoint`, `ref_load` and `save` all point at the checkpoints volume rather than ephemeral disk, and the validation recipe is field-for-field identical to the shipped `000_kimi_k25` tutorial, so a dispatch carries no disk pressure that tutorial doesn't already.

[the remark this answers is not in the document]
zhouhelena1: Checked: `rollout_batch_size=32` on both Kimi recipes, so a one-step validation materializes 32 DAPO-Math-17k prompts — a small download plus a 32-line jsonl write, not a meaningful cost. The scaling is deliberate rather than copied from slime: a rollout step consumes `rollout_batch_size` prompts, so a fixed `n_rows=10` would under-fill even a single step unless the loader wraps epochs.

modal_training_gym/common/models/validation.py
```diff
@@ -9,12 +21,60 @@
...
+class _ValidationConfig:
+    """One model/framework pair the validation harness runs."""
+
+    # Short name used as ``check --model`` and as the CI matrix entry.
+    name: str
+    model_config: type[ModelConfig]
+    # Which framework's ``get_base_recipe`` trains this model.
+    framework: Framework
+    # Whether a pull request may launch this run.
+    ci_enabled: bool = True
```
anli5005: If it's still possible to run these models in CI (which you can via manual dispatch) we should name this something different and more specific

scripts/validate_model_configs.py
```diff
@@ -621,9 +563,21 @@ def __main__():
...
         "list", help="Print available model names as a JSON array and exit."
     )
+    list_parser.add_argument(
+        "--framework",
+        choices=[framework.value for framework in Framework],
+        default=None,
+        help="Only list models validated on this framework.",
+    )
+    list_parser.add_argument(
+        "--all",
```
anli5005: I would argue we need to have the opposite behavior, i.e. a flag that explicitly opts *in* to only showing PR-triggerable models. Otherwise as a dev it's difficult to know that this flag exists without a priori knowledge

zhouhelena1: makes sense to me, changed so list prints the whole registry & --pr-only is the opt-in

modal_training_gym/common/models/validation.py
zhouhelena1: ah ok I had it originally as `gates_prs` but was suggested to change to `ci_enabled`. I agree with the dispatch piece, so I think maybe `run_on_pr` is clearest & most specific?

scripts/validation_backends/miles.py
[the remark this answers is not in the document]
zhouhelena1: checked, slime doesn't carry latent limit, they do differ

scripts/validate_model_configs.py
```diff
@@ -140,170 +164,74 @@ def from_dict(cls, data: dict) -> "TutorialResult":
...
 
-def available_model_names() -> list[str]:
-    """Sorted short model names (e.g. "qwen3-4b") validatable on slime.
+def _ship_dataset_definition(dataset) -> None:
+    """Send the dataset's defining module by value, not by reference.
 
-    The shared registry excludes models with no base slime recipe (e.g. Kimi on
-    miles), since this script only runs base training on slime.
+    ``resolve_caller_context`` only registers the module that calls ``train()``,
+    which leaves a dataset class defined in a backend module pickled by name.
```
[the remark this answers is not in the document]
zhouhelena1: fixed by having download_perf_baseline derive its list from the results directory rather than the registry

joyliu-q (approved):

PR #359, opened 2026-08-08
Measure substep timing in detail

cskitty:
New substep timing implementation! 

Format:
```
patched call site
  → RoleRecorder.phase("forward_backward") 
  → shared background reporting queue (the one rollouts and status use)
  → POST /api/timing-events 
  → substep-timing/<run>/<rollout:08d>__<role>.json   (written during the request)
  → GET /api/runs/<id>/timings
  → one wall-clock timeline, all lanes
```

The unit of storage is a **lane** = one `(rollout_id, role)` pair (`driver` / `actor` / `rollout`, plus a `pre-loop__driver` lane for bootstrap weight sync). Lanes never share a file. Snapshots are cumulative rather than deltas, which is what makes every part of this path idempotent: a skipped, superseded or dropped publish costs staleness, never data, because the next snapshot is a superset of the lost one.

Each phase keeps `count`, `busy_duration_s`, `longest_duration_s`, `first_start_s`, `last_end_s` and the bounded interval list. `busy_duration_s` is the **sum of invocations**, not `last_end_s - first_start_s`, so a phase interleaved with other work doesn't absorb the gaps. Per-sample phases (`reward`, `reward_batch`, `sample_generation`) keep aggregates only — they fire per sample and an interval list would be unbounded in a way the cap can't make meaningful.
 
### Verification runs

*The accuracy runs in this table were measured on this branch's tree; their dashboard pages now render empty because the melody-dev metadata volume was pruned later in development (the run records were deleted, the measurements below stand as recorded).*

Three steps of each mode on this branch's tree, deltas are *measured minus `perf/*`* (these runs predate the `total_duration_s` → `busy_duration_s` rename; both keys still parse):

| Mode | Run | Lanes | `compute_log_probs` vs `perf/log_probs_time` | `forward_backward + optimizer_step` vs `perf/actor_train_time` | `generate_samples` vs `perf/rollout_time` |
|---|---|---|---|---|---|
| slime sync | [`jolly-platform-490e576925b7`](https://modal-labs-melody-dev--training-gym-dashboard-fastapi-app.modal.run/training/jolly-platform-490e576925b7) | driver, actor, rollout, pre-loop | +0.13ms / +0.10ms / +0.19ms | −10.6ms / −10.5ms / −9.9ms | −0.18ms / −0.17ms / −0.16ms |
| slime async | [`blue-vibes-3e64ccc9db7a`](https://modal-labs-melody-dev--training-gym-dashboard-fastapi-app.modal.run/training/blue-vibes-3e64ccc9db7a) | driver, actor, rollout, pre-loop | +0.21ms / +0.11ms / +0.11ms | −12.4ms / −11.6ms / −10.1ms | −0.21ms / −0.15ms / −0.16ms |
| miles sync | [`buttery-bugle-6f689c26a609`](https://modal-labs-melody-dev--training-gym-dashboard-fastapi-app.modal.run/training/buttery-bugle-6f689c26a609) | driver, actor, rollout, pre-loop | +1.27ms / +0.18ms / +0.19ms | −40.9ms / −35.3ms / −35.0ms | −0.05ms / −0.05ms / −0.04ms |
| miles async | [`lime-bud-39de634c5c26`](https://modal-labs-melody-dev--training-gym-dashboard-fastapi-app.modal.run/training/lime-bud-39de634c5c26) | driver, actor, rollout, pre-loop | +0.28ms / +0.13ms / +0.15ms | −36.2ms / −33.1ms / −33.9ms | −0.06ms / −0.06ms / −0.07ms |

 

Link to Devin session: https://modal.devinenterprise.com/sessions/74a52429672c4f2fa48700835f45c0c0
Requested by: @cskitty


dashboards/frontend/src/lib/timing.js
```diff
@@ -0,0 +1,667 @@
...
+
+export const TRAIN_OUTLINE_COLOR = slot("train-outline");
+
+export const CATEGORIES = {
+  train: { label: "Train", color: slot("primary-1") },
+  generate: { label: "Rollout", color: slot("primary-3") },
+  transfer: { label: "Weight sync", color: slot("primary-2") },
+  checkpoint: { label: "Checkpoint", color: slot("primary-5") },
+  eval: { label: "Eval", color: slot("primary-4") },
+  idle: { label: "Waiting", color: "var(--color-c-gray-30)" },
```
joyliu-q: nit: Rename to idle

dashboards/frontend/src/lib/timing.js
```diff
@@ -0,0 +1,667 @@
...
+
+export const TIMING_LABELS = {
+  evaluate_rollouts: "Eval (before training)",
+  evaluate_rollouts_end: "Eval (after training)",
+  generate_rollouts: "Rollout generation",
+  offload_rollout: "Offload generation engines",
+  compute_log_probs: "Calculate log probs",
+  train_models: "Train",
+  training: "Train",
+  train_model: "Train",
```
joyliu-q: why are there so many labels that map to train

joyliu-q: e.g. what is diff between train_models and train_model

dashboards/frontend/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
+    "test": "node --test src/lib/timing.test.js",
```
joyliu-q: Let's hook this into CI?

dashboards/frontend/src/lib/timing.test.js
```diff
@@ -0,0 +1,88 @@
+import assert from "node:assert/strict";
```
joyliu-q, line 1: We don't have any CI to autorun this test, and it relies on the user manually triggering it while in development, which will almost never happen

modal_training_gym/common/run.py
```diff
@@ -113,7 +110,9 @@ class TrainingRun(BaseModel):
     completed_at: int | None = None
     updated_at: int = 0
     duration_seconds: int | None = None
+    # Read-only for legacy substep runs
```
joyliu-q: Maybe add # TODO(joy): Remove this

modal_training_gym/common/run.py
```diff
@@ -113,7 +110,9 @@ class TrainingRun(BaseModel):
...
     step_times: dict[str, dict[str, int | None]] | None = None
+    # Read-only for legacy substep runs
```
joyliu-q: Same here

modal_training_gym/common/status_reporter.py
```diff
@@ -35,6 +35,7 @@
 _LOCK = threading.Lock()
 _DEFAULT_TIMEOUT_SECONDS = 2.0
 _STATUS_TOKEN_ENV = "TRAINING_GYM_FRAMEWORK_STATUS_TOKEN"
+PostResult = Literal["ok", "not_found", "auth_failed", "failed", "permanent"]
```
joyliu-q: This should be an `enum.Enum`

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,91 @@
...
+def is_safe_run_id(value: str) -> bool:
+    return (
+        bool(value)
+        and value[0] != "."
+        and all(
+            character
+            in "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-"
+            for character in value
+        )
+    )
```
joyliu-q: When would the run id, which is something we control the generation of, become unsafe?

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,91 @@
...
+
+
+class Role(str, Enum):
+    DRIVER = "driver"
+    ROLLOUT = "rollout"
+    ACTOR = "actor"
+    CRITIC = "critic"
```
joyliu-q, line 36: What is a critic?

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,91 @@
...
+        and all(
+            character
+            in "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-"
+            for character in value
+        )
+    )
+
+
+class Role(str, Enum):
+    DRIVER = "driver"
```
joyliu-q, line 29: Maybe add comments on what these roles are since both me and Shanka didn't know what a driver was

Same thing w Ui, showing on hover an explanation or adding a legend.

modal_training_gym/common/run.py
cskitty: If these are removed I don't think runs done while the old step timing will have step times viewable anymore - but seems like something that should/could be added when a long enough time has passed? added.

modal_training_gym/common/step_timing.py
cskitty: added in comments

modal_training_gym/common/step_timing.py
cskitty: added in UI

modal_training_gym/common/step_timing.py
cskitty: Run id is used in the get request - /api/runs/{id}/rollouts which is unauthenticated, my original thought was that if this data should be private then safe run id could guard against people incorrectly trying to hit this endpoint. But actually if they had the run id this wouldn't do anything anyway so will remove

dashboards/frontend/src/lib/timing.js
cskitty: this was unnecessary, removed

dashboards/frontend/src/lib/timing.js
cskitty: added

modal_training_gym/common/reporting.py
```diff
@@ -157,6 +157,13 @@ def _enqueue(payload: dict[str, Any]) -> None:
         pass
 
 
+def _post_framework_status(payload: dict[str, Any], timeout: float) -> None:
+    url = _phase_url()
+    if not url:
+        return
+    _post({"_url": url, "_timeout": timeout, **payload})
```
joyliu-q: nit: Not needed diff, just moves the file

modal_training_gym/common/reporting.py
```diff
@@ -157,6 +157,13 @@ def _enqueue(payload: dict[str, Any]) -> None:
...
+
```
joyliu-q: ditto

modal_training_gym/common/status_reporter.py
```diff
@@ -37,6 +38,14 @@
 _STATUS_TOKEN_ENV = "TRAINING_GYM_FRAMEWORK_STATUS_TOKEN"
 
 
+class PostResult(str, Enum):
+    OK = "ok"
+    NOT_FOUND = "not_found"
+    AUTH_FAILED = "auth_failed"
+    FAILED = "failed"
+    PERMANENT = "permanent"
```
joyliu-q: Would prefer `_PostResult` with `OK` vs `FAILED`, and for failed there is a separate `_PostResultFailure` enum.

Also since this is internal we should underscore all internal classes.

What is PERMANENT and NOT_FOUND, and when are they in action

modal_training_gym/common/status_reporter.py
```diff
@@ -108,8 +117,19 @@ def _post(item: dict[str, Any]) -> None:
     try:
         with urlopen(request, timeout=timeout) as response:
             response.read()
+    except HTTPError as exc:
+        if exc.code in {404, 405}:
+            return PostResult.NOT_FOUND
```
joyliu-q: When would this happen

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,97 @@
...
+        bool(value)
+        and value[0] != "."
+        and all(
+            character
+            in "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-"
+            for character in value
+        )
+    )
+
+
```
joyliu-q: ?

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,97 @@
...
+    ACTOR = "actor"
+    # The value model that scores states, present only in PPO-style runs
+    # (use_critic=true); absent for GRPO.
+    CRITIC = "critic"
+
+
+class PhaseTiming(BaseModel):
+    count: int
+    total_duration_s: float
+    longest_duration_s: float
```
joyliu-q: What is longest_duration_s?

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,97 @@
...
+    first_start_s: float
+    last_end_s: float
```
joyliu-q, line 49: Cant u derive total duration by using `first_start_s` and `last_end_s`? It feels like you are having redundant information.

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,97 @@
...
+    @property
+    def average_duration_s(self) -> float:
+        return self.total_duration_s / self.count if self.count else 0.0
+
+
+class RoleTimingRecord(BaseModel):
+    training_run_id: str
+    rollout_id: int | None = Field(default=None, ge=0)
+    role: Role
+    created_at: int = 0
```
joyliu-q: When is created_at set

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,97 @@
...
+    lane_start_unix_s: float | None = None
```
joyliu-q, line 59: unix?

modal_training_gym/common/step_timing.py
joyliu-q: If something has `_at`, it should be a `datetime`

modal_training_gym/common/step_timing.py
```diff
@@ -1,10 +1,97 @@
...
+        rollout = "pre-loop" if self.rollout_id is None else f"{self.rollout_id:08d}"
+        return f"{rollout}__{self.role.value}"
+
+    @staticmethod
+    def store(training_run_id: str) -> str:
+        return f"{MetadataStore.SUBSTEP_TIMING.value}/{training_run_id}"
+
+    def _touch_created_at(self) -> None:
+        if not self.created_at:
+            self.created_at = int(time.time())
```
joyliu-q: This function is only called once and is unneeded

modal_training_gym/common/step_timing.py
```diff
@@ -18,58 +105,135 @@ class Substep(str, Enum):
...
+
+
+def _rollout_sort_key(item: tuple[int | None, Any]) -> tuple[bool, int]:
+    rollout_id = item[0]
+    return rollout_id is None, 0 if rollout_id is None else rollout_id
+
+
+_LEGACY_RENAMES = {Substep.OPTIMIZER_STEP.value: Substep.TRAIN_MODELS.value}
+
+
```
joyliu-q, line 232: When will this then be deprecated

modal_training_gym/frameworks/miles/modal_helpers/patches/patch_sglang_abort.py
```diff
@@ -43,4 +43,5 @@
     path.write_text(src.replace(old, new, 1))
     print("Patched Miles SGLang abort cleanup")
 else:
-    raise RuntimeError("Could not find Miles SGLang abort block to patch")
+    message = "Could not find Miles SGLang abort block to patch"
+    raise RuntimeError(message)
```
joyliu-q: Unnecessary diff

modal_training_gym/utils/metadata.py
```diff
@@ -153,6 +150,28 @@ def _store_path(store: MetadataStore | str) -> str:
...
+    readers: Iterable[Callable[[], Awaitable[T]]],
+) -> list[T | BaseException]:
+    semaphore = asyncio.Semaphore(16)
+
+    async def _read(reader: Callable[[], Awaitable[T]]) -> T:
+        async with semaphore:
+            for attempt in range(3):
+                try:
+                    return await reader()
+                except Exception as exc:
```
joyliu-q: Generic `Exception`, what specific exceptions do we expect?

modal_training_gym/utils/metadata.py
```diff
@@ -226,24 +245,60 @@ async def _run() -> dict[str, Any]:
...
+                        entry.path
+                        async for entry in vol.iterdir.aio(_store_path(store))
+                        if entry.path.endswith(".json")
+                    ]
+                    break
+                except (FileNotFoundError, NotFoundError):
+                    if return_failures:
+                        return [], False
+                    return []
+                except Exception as exc:
```
joyliu-q: ditto, there needs to be sufficient reason for us to add this catchall

modal_training_gym/utils/metadata.py
```diff
@@ -257,16 +312,77 @@ async def _run() -> list[dict[str, Any]]:
...
+def vol_list_metadata(
+    store: MetadataStore | str,
+    *,
+    is_async: bool = False,
+    return_failures: bool = False,
+) -> (
+    list[dict[str, Any]]
+    | tuple[list[dict[str, Any]], bool]
+    | Awaitable[list[dict[str, Any]] | tuple[list[dict[str, Any]], bool]]
+):
```
joyliu-q: Do we not already have this?

modal_training_gym/common/status_reporter.py
```diff
@@ -125,10 +152,14 @@ def enqueue_item(item: dict[str, Any]) -> None:
         pass
 
 
-def post_item(item: dict[str, Any]) -> None:
-    """Synchronously POST a pre-resolved item (same shape as ``enqueue_item``),
-    blocking up to the item's ``_timeout``. Failures are swallowed."""
-    _post(item)
+def post_item(item: dict[str, Any]) -> bool:
+    return _post(item)[0] == _PostResult.OK
```
joyliu-q: What's the point of categorizing the error if we are just going to swallow the error type (second field) up without logging?

modal_training_gym/common/status_reporter.py
```diff
@@ -125,10 +152,14 @@ def enqueue_item(item: dict[str, Any]) -> None:
...
+
+
+def post_item_result(
+    item: dict[str, Any],
+) -> tuple[_PostResult, _PostResultFailure | None]:
+    return _post(item)
```
joyliu-q: Why two diff functions here? Also `post_item_result` is the exact same as `_post`

modal_training_gym/common/step_timing.py
```diff
@@ -18,58 +89,142 @@ class Substep(str, Enum):
...
+        is_async=True,
+    )
+    records = [
+        RoleTimingRecord.model_validate(record).model_dump(mode="json")
+        for record in records
+    ]
+    grouped: dict[int | None, list[dict[str, Any]]] = {}
+    for record in records:
+        grouped.setdefault(record["rollout_id"], []).append(record)
+    return grouped, had_failures
```
joyliu-q: Make it a single `load_run` function that takes in `is_async` as a parameter, similar to `vol_list_with_failures`

modal_training_gym/common/status_reporter.py
cskitty: Deleted

modal_training_gym/common/step_timing.py
cskitty: Done

modal_training_gym/common/status_reporter.py
cskitty: Deleted

modal_training_gym/common/step_timing.py
cskitty: renamed total duration to busy duration, it's only actively used time

modal_training_gym/common/status_reporter.py
```diff
@@ -73,12 +86,14 @@ def _worker() -> None:
         if item is None:
             return
         try:
-            _post(item)
+            post_item(item)
```
joyliu-q: Why are we not logging if it fails? `post_item` returns some values but it is repressed atm

joyliu-q (approved):
:0

PR #363, opened 2026-08-10
Add async Miles smoke script for Qwen3-0.6B

cskitty:
Adds `scripts/run_async_miles_smoke.py`: the MILES analogue of `validate_model_configs.run_base_training_on_slime` (same gsm8k prompts, `rm_type="deepscaler"`), but with `MilesRecipe(async_mode=True)` so rollout generation overlaps training. Two rollouts, `rollout_batch_size=8`, `n_samples_per_prompt=4`, 512-token responses.

Two constraints fix the topology at 2×8×H100, which the comments in the file record:
- miles `train_async.py` asserts `not args.colocate`, so async can't run colocated (the 1-node colocated attempt failed on that assertion).
- Modal clustered functions must take a whole node's GPUs, so `actor_num_gpus_per_node=1` is rejected (`InvalidError: ... requires 8`).

The gsm8k dataset class is defined in the script rather than imported from `validate_model_configs`; imported it deserialized remotely as `ModuleNotFoundError: No module named 'validate_model_configs'` even with `cloudpickle.register_pickle_by_value`.

Verified run: `pale-deadline-065708c79d2c` — COMPLETED in 567s, `rollout 0` / `rollout 1` logged with rewards and two `train_step ... outcome=NORMAL valid_step=true` steps.

## Checklist

Link to Devin session: https://modal.devinenterprise.com/sessions/7288eb27e40d434bab80b0b60bc66116
Requested by: @cskitty

### DIFF
--- a/scripts/run_async_miles_smoke.py
+++ b/scripts/run_async_miles_smoke.py
@@ -1,0 +1,76 @@
+"""Minimal async MILES smoke run on Qwen3-0.6B.
+
+Mirrors the slime validator (``validate_model_configs.run_base_training_on_slime``):
+same gsm8k dataset and ``rm_type="deepscaler"``, but on ``MilesRecipe`` with
+``async_mode=True`` so rollout generation overlaps training. The goal is to
+prove the async rollout path works end-to-end, not to train a useful model.
+
+    uv run scripts/run_async_miles_smoke.py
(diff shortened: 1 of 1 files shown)


cskitty:
not intended for pr

PR #355, opened 2026-08-07
Add basic API for creating Modal endpoints from the gym

anli5005:
Eventually, we want to migrate `DeploymentConfig` to a Modal Endpoints-based API. This PR sets up the first part of that, by adding an API that allows users to create endpoints from existing ModelConfigs. The underlying implementation borrows heavily from @andrewhinh's work on #327.

Example of what the current API looks like:
```python
model = Qwen3_6_27B()
endpoint = Endpoint.launch(model, unauthenticated=False)
endpoint.wait_until_ready()
endpoint.chat(messages=[
    {
        "role": "system",
        "content": "You are an expert web developer.",
    },
    {
        "role": "user",
        "content": "In a single HTML file, write a web page that helps me track porcupines I've seen, saving the data to IndexedDB.",
    },
])
```

The old `DeploymentConfig` equivalent is as follows:
```python
deployment = DeploymentConfig(
    model=model,
    unauthenticated=False,
).serve()
deployment.wait_until_ready()
deployment.chat(messages=[
    {
        "role": "system",
        "content": "You are an expert web developer.",
    },
    {
        "role": "user",
        "content": "In a single HTML file, write a web page that helps me track porcupines I've seen, saving the data to IndexedDB.",
    },
])
```

~~I've intentionally withheld checkpoint support from this initial PR; that will likely come in a later PR once we figure out overall direction.~~ Checkpoint support has been added via a new `checkpoint=` parameter! However, any Megatron checkpoints will need to be converted to HF format, like so:

```python
from modal_training_gym.common.checkpoint import convert_checkpoint_to_hf
converted = convert_checkpoint_to_hf(checkpoint, model)
endpoint = Endpoint.launch(model, checkpoint)
```

PR #372, opened 2026-08-11
Consolidate DeploymentConfig and ModelDeployment into CustomDeployment, matching Endpoint

anli5005:
This PR consolidates our existing `DeploymentConfig` and `ModelDeployment` APIs, which manually spin up a vLLM or SGLang server on Modal, into a single `CustomDeployment` class that loosely matches the `Endpoint` API introduced in #355.

This will facilitate migrations to Modal Endpoints as more tutorials are updated, in addition to removing redundancy and making the API slightly easier to reason about.

Before:

```python
deployment: ModelDeployment = DeploymentConfig(
    model=Qwen3_4B(),
    checkpoint=checkpoint,
    recipe=SglangRecipe(gpu="H100"),
    app_name="qwen3-serve",
).serve()
```

After:

```python
deployment: CustomDeployment = CustomDeployment.launch(
    Qwen3_4B(),
    checkpoint,
    recipe=SglangRecipe(gpu="H100"),
    app_name="qwen3-serve",
)
```

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/372" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


modal_training_gym/common/deployment.py
```diff
@@ -401,60 +222,189 @@ def observe(self, containers: int, now: float) -> tuple[int, str] | None:
...
 
 
-class ModelDeployment(BaseModel):
-    """A deployed model endpoint.
+class AdHocDeployment(BaseModel):
+    """A recipe-driven deployed model endpoint.
 
-    Returned by ``DeploymentConfig.serve()``. Carries the live endpoint
-    URL, Modal app id, and convenience methods for generation and evaluation.
+    Use :meth:`launch` to build and deploy an SGLang or vLLM serving app.
```
andrewhinh: lmao what is :meth:

modal_training_gym/common/deployment.py
```diff
@@ -401,60 +222,189 @@ def observe(self, containers: int, now: float) -> tuple[int, str] | None:
...
+        cls,
+        model: ModelConfig | str,
+        checkpoint: Checkpoint | None = None,
+        *,
+        recipe: VllmRecipe | SglangRecipe,
+        app_name: str | None = None,
+        served_model_name: str | None = None,
+        unauthenticated: bool = True,
+    ) -> "AdHocDeployment":
+        """Build, deploy, and return a recipe-driven deployment handle."""
```
andrewhinh:
```suggestion
```

modal_training_gym/common/deployment.py
```diff
@@ -401,60 +222,189 @@ def observe(self, containers: int, now: float) -> tuple[int, str] | None:
...
     modal_app_url: str = ""
     url: str
     status: str = DeploymentStatus.RUNNING.value
 
-    @field_validator("deployment_config", mode="before")
+    @model_validator(mode="before")
     @classmethod
-    def _parse_deployment_config(cls, value: object) -> DeploymentConfig | object:
-        if not isinstance(value, dict):
+    def _parse_legacy_deployment(cls, value: object) -> object:
```
andrewhinh: why is this needed?

modal_training_gym/common/deployment.py
```diff
@@ -401,60 +222,189 @@ def observe(self, containers: int, now: float) -> tuple[int, str] | None:
...
+                checkpoints_mount_path=checkpoints_mount_path,
+                deployment_id=deployment_id,
+                unauthenticated=unauthenticated,
+            )
+        elif isinstance(recipe, VllmRecipe):
+            if unauthenticated is False:
+                warnings.warn(
+                    "AdHocDeployment.launch(..., unauthenticated=False) is not "
+                    "supported for VllmRecipe: "
+                    "modal.experimental.http_server has no proxy-auth, so the "
```
andrewhinh: aren't we using `app.server` everywhere now?

modal_training_gym/common/train_result.py
```diff
@@ -23,8 +23,9 @@
     print(result.checkpoint_dir)            # /checkpoints/my-app_train_...
     print(result.latest_checkpoint_path())  # .../iter_0000050
 
-    from modal_training_gym.common.deployment import DeploymentConfig
-    deployment = DeploymentConfig(model=result.model).serve()
+    from modal_training_gym import AdHocDeployment
+    from modal_training_gym.deploy_recipes import SglangRecipe
+    deployment = AdHocDeployment.launch(result.model, recipe=SglangRecipe())
```
andrewhinh: why are we passing `recipe` here when before we didn't have to?

andrewhinh (changes requested):

joyliu-q:
Can we not call it AdHocDeployment but `Deployment` or `CustomDeployment`

andrewhinh:
+1 on `CustomDeployment`

modal_training_gym/common/deployment.py
anli5005: It's a [Sphinx directive to cross-reference methods](https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-meth) which we've used a few times in the environments codebase:

https://github.com/modal-projects/training-gym/blob/fb3b00fdd6b7617561d0e853ba2df7d830ad3b05/modal_training_gym/common/environments/base.py#L56-L58

Unfortunately we don't actually seem to parse this in our docs pipeline, so will remove.

modal_training_gym/common/deployment.py
anli5005: We persist these to disk as Pydantic models, so we need to run some validation to adapt from the legacy format to the new one where `deployment_config` no longer exists.

I'll rename this to `_parse_deployment` though since the "legacy" moniker is a bit heavy-handed

modal_training_gym/common/train_result.py
anli5005: The API changed to require a `recipe`, but I'll change it back to default to an `SglangRecipe()`.

modal_training_gym/common/deployment.py
anli5005: We are, it just looks like we never threaded the `unauthenticated` flag on the vLLM side after the migration. Will do that rq

modal_training_gym/cli/setup.py
```diff
@@ -128,7 +128,8 @@ def ensure_proxy_auth(interactive: bool = True, force: bool = False) -> bool:
             return True
     else:
         print(
-            "\nAuthenticated served endpoints (DeploymentConfig(unauthenticated=False)) need "
+            "\nAuthenticated ad-hoc deployments "
```
andrewhinh:
```suggestion
            "\nAuthenticated custom deployments "
```

ad-hoc should be replaced with custom everywhere.

tutorials/tutorial_generator/agent/000_agent_sandbox.py
```diff
@@ -183,11 +182,11 @@ def _deploy_model():
     recipe = SglangRecipe(
         extra_server_args={"--tool-call-parser": "qwen25"},
     )
-    deployment = DeploymentConfig(
-        model=Qwen3_8B(),
+    deployment = CustomDeployment.launch(
+        Qwen3_8B(),
```
andrewhinh, line 186: i feel like we should still specify `model=`; wdyt?

tutorials/tutorial_generator/agent/000_agent_sandbox.py
```diff
@@ -166,8 +165,8 @@ def _deploy_section():
     """
     ## Deploy the model
 
-    `DeploymentConfig.serve()` launches an sglang-backed inference
-    server on Modal and returns a `ModelDeployment` with a live URL.
+    `CustomDeployment.launch()` launches an sglang-backed inference
+    server on Modal and returns an `CustomDeployment` with a live URL.
```
andrewhinh:
```suggestion
    server on Modal and returns a `CustomDeployment` with a live URL.
```

should be fixed throughout.

andrewhinh (changes requested):

andrewhinh (approved):

PR #377, opened 2026-08-12
Rework dataset API

anli5005:
This PR significantly reworks the `DatasetConfig` API, with the goal of making it more understandable, maintainable, and robust. Previously, the API looked something like this:

```python
class DatasetConfig:
    # Return an iterable collection of rows
    def load(split) -> list[DatasetRow]

    # Write the test and eval datasets to the specified paths
    def prepare(path, eval_paths) -> None
```

Now, it looks like this:

```python
class DatasetConfig(ABC):
    def rows() -> Iterable[DatasetRow]
```

Other notable changes:
* Datasets are no longer expected to return both a train and eval split; instead, **evaluation datasets are now instantiated separately** ~and passed into TrainConfig as a separate, optional eval_dataset parameter.~ Previously, different datasets treated their eval split in inconsistent ways; for example, `HuggingFaceDataset` returned the same data for train/eval, while other datasets attempted to do their own splitting. This change gives significantly more control over the built-in evaluations Slime and Miles run.
* Materializing a dataset to disk is now handled by a `DatasetConfig.write` method. While we provide a default implementation that iterates over `rows()` and writes them to a JSONL file, subclasses such as `HuggingFaceDataset` can override this to implement optimizations or add support for Parquet.
* ~Many config flags stored as fields have been migrated to `@property` getters. While this is less concise, it makes it easier for subclasses to dynamically change behavior depending on the data they contain.~ Some flags have been renamed for clarity, e.g. `apply_chat_template` is now `needs_chat_template`, and `always_prepare` is now `requires_prepare_before_train`.
* Dataset initializers no longer take arbitrary keyword arguments.

This is a fairly wide-reaching and breaking change, affecting all existing dataset implementations and tutorials.

### DIFF
--- a/modal_training_gym/common/dataset.py
+++ b/modal_training_gym/common/dataset.py
@@ -60,7 +60,7 @@ class DatasetConfig(ABC):
         return True
 
     @property
-    def requires_refresh_before_training(self) -> bool:
+    def needs_refresh(self) -> bool:
         return False
 
     @abstractmethod
@@ -292,7 +292,7 @@ class HarborDataset(DatasetConfig):
         eval_repeats: int | None = None,
         shuffle_tasks: bool | None = None,
         shuffle_seed: int | None = None,
-        requires_refresh_before_training: bool = False,
+        needs_refresh: bool = False,
     ) -> None:
         self.split = split
         # Keep this while framework path resolution still uses ``hf_split``.
@@ -329,7 +329,7 @@ class HarborDataset(DatasetConfig):
             self.shuffle_tasks = shuffle_tasks
         if shuffle_seed is not None:
             self.shuffle_seed = shuffle_seed
-        self._requires_refresh_before_training = requires_refresh_before_training
+        self._needs_refresh = needs_refresh
         if not self.id:
             self.id = f"{self._id_slug()}-{split}-{uuid.uuid4()}"
         super().__init__()
@@ -360,8 +360,8 @@ class HarborDataset(DatasetConfig):
         return "parquet"
 
     @property
-    def requires_refresh_before_training(self) -> bool:
-        return self._requires_refresh_before_training
+    def needs_refresh(self) -> bool:
+        return self._needs_refresh
 
     def _harbor_dataset_ref(self) -> str:
         if "@" in self.dataset_name:
--- a/modal_training_gym/common/dataset.py
+++ b/modal_training_gym/common/dataset.py
@@ -64,7 +64,7 @@ class DatasetConfig(ABC):
     input_key: str = "input"
     needs_chat_template: bool = True
     needs_refresh: bool = False
-    writes_eval_paths: bool = True
+    writes_eval_paths: bool = False
     output_format: Literal["parquet", "jsonl"] = "parquet"
 
     @property
@@ -186,6 +186,7 @@ class HuggingFaceDataset(DatasetConfig):
     """
 
     type: DatasetType = DatasetType.HUGGING_FACE
+    writes_eval_paths: bool = True
     hf_repo: str = ""
     hf_split: str = "train"
     hf_config: str | None = None
@@ -289,6 +290,7 @@ class HarborDataset(DatasetConfig):
     """
 
     type: DatasetType = DatasetType.HARBOR
+    writes_eval_paths: bool = True
     dataset_name: str = ""
     path: str | None = None
     task_root: str = ""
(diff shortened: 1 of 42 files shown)


modal_training_gym/common/environments/bfcl.py
```diff
@@ -588,26 +590,34 @@ def _category_filename(category: str) -> str:
 class BfclMultiTurnDataset(DatasetConfig):
     """Loads a BFCL multi-turn category from the installed ``bfcl_eval`` package.
 
-    Last :attr:`config`.eval_tail ids are held out as eval.
+    The last :attr:`config`.eval_tail ids are held out for an instance configured
```
andrewhinh, line 593:
```suggestion
    The last `config`.eval_tail ids are held out for an instance configured
```

These aren't rendered iirc

modal_training_gym/common/dataset.py
```diff
@@ -30,56 +32,52 @@ class DatasetType(Enum):
...
 
-    def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
-        """Materialize training data to ``path`` (and eval splits to ``eval_paths``)."""
-        raise NotImplementedError(f"{type(self).__name__} has no prepare()")
+    @property
+    def needs_chat_template(self) -> bool:
+        return True
+
+    @property
+    def requires_refresh_before_training(self) -> bool:
```
andrewhinh:
```suggestion
    def needs_refresh(self) -> bool:
```

modal_training_gym/common/deployment.py
```diff
@@ -309,6 +309,7 @@ def launch(
             checkpoint.path if checkpoint is not None else "",
             f"{type(recipe).__name__}:{recipe.recipe_type.value}",
             app_name,
+            "",
```
andrewhinh: can we do something else than null?

scripts/validate_model_configs.py
```diff
@@ -264,6 +265,7 @@ def run_base_training(
     train_config = TrainConfig(
         model=model_config,
         dataset=dataset,
+        eval_dataset=copy.deepcopy(dataset),
```
andrewhinh: why aren't we splitting here?

tutorials/tutorial_generator/rl/003_on_policy_distillation.py
```diff
@@ -180,20 +180,19 @@ def _dataset_intro():
...
+    train_dataset = MathDataset(hf_split="train[:100]")
+    eval_dataset = MathDataset(hf_split="train[100:120]")
 
 
 @notebook_only
 @code
 def _dataset_peek():
-    rows = eval_dataset.load()
-    for row in rows.select(range(2)):
+    for row in list(eval_dataset.rows())[:2]:
```
andrewhinh: why are we materializing all rows if we only intend to use the first 2?

andrewhinh (changes requested):

scripts/validate_model_configs.py
anli5005: I think we should actually set this to `None`, not sure what you think?

modal_training_gym/train_recipes/base.py
```diff
@@ -104,33 +104,30 @@ def validate_model_parallelism(self, model: "ModelConfig") -> None:
...
-        split = getattr(ds, "hf_split", "train")
-        prompt_data = f"{DATA_PATH}/{name}/{split}.{ext}"
-        if getattr(ds, "writes_eval_paths", True):
-            return prompt_data, {"eval": f"{DATA_PATH}/{name}/eval.{ext}"}
-        return prompt_data, None
+        fmt = ds.output_format
+        ext = "jsonl" if fmt == "jsonl" else "parquet"
+        cache_key = getattr(ds, "cache_key", "")
+        suffix = f"-{cache_key}" if cache_key else ""
+        return f"{DATA_PATH}/{name}/{split}{suffix}.{ext}"
```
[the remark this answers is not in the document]
andrewhinh: fixed in [09c3916](https://github.com/modal-projects/training-gym/pull/377/commits/09c39160c3e70f13724272a6f3aa158b7235ee59)

modal_training_gym/common/dataset.py
```diff
@@ -105,8 +105,22 @@ def _expected_columns(self) -> set[str]:
...
+        else:
+            with open(path, "w") as f:
+                for row in rows:
+                    f.write(json.dumps(row) + "\n")
+
+    def write(self, path: str) -> None:
+        self._write_rows(self.rows(), path)
+
+    def write_eval(self, path: str) -> None:
+        self.write(path)
```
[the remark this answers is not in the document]
andrewhinh: Fixed in c0da7eaa053b1e8584cee699f44db1e055178cd7. `DatasetConfig.writes_eval_paths` now defaults to `False`. Only `HuggingFaceDataset` and `HarborDataset` set it to `True`. NumberGuess therefore does not invent eval data. `multiturn.py` uses `Qwen3_5_4b_Recipe(eval_interval=None)`, so Slime does not schedule eval without files. Recipe merging that swallows generic `SlimeRecipe(eval_interval=None)` is owned by PR 445, not this PR.

anli5005:
Closing in favor of #483

PR #379, opened 2026-08-12
Deduplicate rollout input images by content

zhouhelena1:
This PR deduplicates rollout images for samples with same prompt. Previously, input images were inlined per sample under a per-sample cap that counted multiple copies of the same screenshot. Now, RolloutImageStore caps distinct images instead. No visual change on the dashboard.

Split out of #274 (Gemma-4) at review request, so the sampling change lands independently of the model recipe and unblocks #347. 

Screenshot 2026-08-13 at 10 38 12 AM

## Checklist

## Outside contributors

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/379" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


modal_training_gym/common/sample_extraction.py
```diff
@@ -563,12 +695,21 @@ def get(key: str, default: Any = None) -> Any:
...
+    prompt_text = _coerce_text(prompt)
     if audio_uri := _extract_audio_from_prompt(prompt):
         metadata["_metadata_type"] = "audio"
         metadata["audio"] = audio_uri
-    elif include_image and (image_uri := _extract_image_from_sample(sample)):
+    elif image_store is not None and image_store.annotate(
+        sample,
+        metadata,
+        group_key=None if group_index is None else (group_index, prompt_text),
+    ):
```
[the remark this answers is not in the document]
zhouhelena1: Real bug, and worse than the write-up suggests — fixed in bfe3d89b by deleting the cache.

I went looking for the per-turn discriminator you suggested and there isn't one. Prompt-group members and episode turns are indistinguishable by the fields a Sample exposes: both share `group_index`, and both differ in `index`, so `index`/turn-index separates the turns but also splits the group members the cache exists to collapse. Response length is content, not identity.

Which makes the cache unfixable rather than mis-keyed, so it's gone. `_ref_by_raw` content keying was already what decided equality; it now decides it for every sample. The `_key_by_id` memo stays, so a group that shares one image object is still content-hashed once — the case the cache was actually buying anything for. The cost is one `tobytes()` + md5 per sample when a framework hands out distinct copies, which is cheap next to the thumbnail + base64 encode it still avoids.

Both consequences you listed are gone. Verified against a stubbed encoder:

| case | before | after |
|---|---|---|
| 8 equal-content copies (one group) | 1 encode / 1 carrier / 1 ref | unchanged |
| 3 turns, distinct frames, constant prompt | 1 ref, turns 2-3 show frame 1 | 3 distinct refs |
| imageless turn 1, then a frame | later turns denied | turn 2 annotated |
| cap=2 over 4 distinct images | count=2 | unchanged |

On the premise itself: you were right not to take the docstring's word for it. It asserted prompt text varies per turn, which is true for the producers I had in mind and not something the store can rely on — `_compact_trajectory_messages` existing is the tell, as you say.

modal_training_gym/common/sample_extraction.py
```diff
@@ -563,12 +695,21 @@ def get(key: str, default: Any = None) -> Any:
             if compact:
                 metadata["eval_report"] = compact
 
+    sample_index = optional_int(get("index"))
+    group_index = optional_int(get("group_index"))
+    if group_index is None and sample_index is not None:
+        group_index = sample_index // max(1, int(n_samples_per_prompt or 1))
```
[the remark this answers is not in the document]
zhouhelena1: Checked, and no change needed — but the divergence you spotted in the advantage path is real and worth recording, so thanks for it.

On the semantics: `group_index` is a group id, not a within-group position. Verified by reading both frameworks rather than inferring it, at the commits behind the pinned images:

```
# slime/rollout/data_source.py @ a4b8c9c8e2f4
 55:  self.sample_group_index = 0
112:  sample.group_index = self.sample_group_index
113:  sample.index = self.sample_index
114:  self.sample_index += 1
116:  self.sample_group_index += 1
```

miles has the identical shape at `miles/rollout/data_source.py:53,113,114,115,117`. It's a per-prompt counter, so preferring it is strictly better than deriving it — your observation that nothing in *this repo* references the field is accurate, which is exactly why I went to the upstream source.

On `reporting.py:267` disagreeing: that's a pre-existing gap this PR narrows rather than opens. `_build_advantage_rows` takes parallel numeric arrays, not Samples, so it has no `group_index` to prefer — it can only derive positionally. Where the derivation is right the two payloads agree; where it's wrong the sample payload is now correct and the advantage payload is as wrong as it always was. Fixing the other half means threading group ids into `_build_advantage_rows`, which touches a different payload and belongs in its own change — filed as follow-up rather than smuggled in here.

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -396,7 +433,9 @@
       rollouts: groups.length,
       n_samples_per_prompt: expandedRollout.n_samples_per_prompt ?? null,
       mean: scores.reduce((a, v) => a + v, 0) / scores.length,
-      samples: samples.map(sampleToPayload),
+      // Shared screenshots, keyed by the `metadata.image_ref` each sample carries.
```
anli5005: screenshots?

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -350,7 +369,21 @@
...
 
-  function sampleToPayload(s) {
+  function sampleToPayload(s, { inlineImage = false, refOnly = false } = {}) {
+    let metadata = s.metadata || null;
+    if (inlineImage && metadata?.image_ref && !metadata.image) {
+      // Only trade the ref for bytes if the lookup resolved — the carrier sample may
+      // not be loaded, and the download still needs some way to name the screenshot.
+      const resolved = sampleImage(s);
+      if (resolved) {
+        const { image_ref, ...rest } = metadata;
```
anli5005: Is there a reason to strip out the `image_ref` here?

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
```diff
@@ -350,7 +369,21 @@
     }
   }
 
-  function sampleToPayload(s) {
+  function sampleToPayload(s, { inlineImage = false, refOnly = false } = {}) {
```
anli5005: Would prefer a more clear set of parameters since it seems like `{ inlineImage = true, refOnly = true }` doesn't make sense here - maybe something like an `imageHandling` parameter that can be set to `ignore`, `refs_only`, or `resolve`?

anli5005 (commented):
A few nits:

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
zhouhelena1: I like that, will do that ty!

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
zhouhelena1: Oops that's what the tutorial / screenspot dataset uses as images, I'll reword as "Shared images"

dashboards/frontend/src/pages/TrainingRunDetailPage.svelte
zhouhelena1: no reason, I can un-strip it

modal_training_gym/common/sample_extraction.py
```diff
@@ -483,19 +486,127 @@ def get(key: str, default: Any = None) -> Any:
...
+            key = self._key(candidate)
+            cached = self._ref_by_raw.get(key)
+            if cached is not None:
+                if not cached:
+                    continue
+                metadata["image_ref"] = cached
+                return cached
+            if self.count >= self._limit:
+                self._ref_by_raw[key] = ""
+                continue
```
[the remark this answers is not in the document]
anli5005: Ah yes this is a good point

anli5005 (approved):

anli5005:
Will approve for now, feel free to merge once the Devin comment is addressed

modal_training_gym/common/sample_extraction.py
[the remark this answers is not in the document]
zhouhelena1: fixed

modal_training_gym/common/sample_extraction.py
```diff
@@ -483,19 +486,137 @@ def get(key: str, default: Any = None) -> Any:
...
+            if not uri:
+                self._ref_by_raw[key] = ""
+                continue
+            ref = self._ref_by_uri.get(uri)
+            if ref is None:
+                ref = hashlib.md5(uri.encode()).hexdigest()[:_IMAGE_REF_CHARS]
+                self._ref_by_uri[uri] = ref
+                metadata["image"] = uri
+            self._ref_by_raw[key] = ref
+            metadata["image_ref"] = ref
```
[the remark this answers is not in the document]
zhouhelena1: fixed

PR #402, opened 2026-08-15
Honor TRAINING_GYM_FRAMEWORK_STATUS_URL in get_framework_status_url

pedramr:
`status_reporter._resolve_url()` already honors `TRAINING_GYM_FRAMEWORK_STATUS_URL`, but `get_framework_status_url()` does not. `train.py` resolves the status URL through `get_framework_status_url()` on the driver and passes it into the workers, so when the env var is set the driver and the reporters disagree on the endpoint.

This moves the override into `get_framework_status_url()` and has `_resolve_url()` delegate to it, so the precedence lives in one place.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/402" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/config.py
+++ b/modal_training_gym/common/config.py
@@ -182,7 +182,15 @@ def modal_proxy_auth_headers() -> dict[str, str]:
 
 
 def get_framework_status_url() -> str | None:
-    """Return the saved framework-status endpoint URL, or ``None``."""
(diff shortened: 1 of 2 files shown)


anli5005 (approved):
Looks good, thanks for the contribution!

PR #404, opened 2026-08-16
Name tracker runs after the training run

pedramr:
`wandb_run_id_for_attempt()` returns `training_run_id[:8]`. W&B validates run
ids for charset only (`/\#?%:`) with no length cap, and these are `[a-z0-9-]`,
so pass the whole thing.

The full id travels better across wandb-compatible backends: W&B keeps a display
name separate from the id, so truncation only ever shows in its URLs, but
trackio and friends use the id *as* the run name.

It also removes a collision. A training run id is a random slug plus a 12-hex
digest (`create_hash`), so 8 characters is mostly slug and drops what
distinguishes two runs:

```
electric-batter-6362579afd91      -> electric
electrical-solution-5b54ee2f4ca8  -> electric
```

Over `randomname`'s vocabulary, 5000 ids give 4993 distinct slugs but only 2814
distinct 8-char prefixes — median 51 runs to a project's first collision. The
launchers export this as `WANDB_RUN_ID` with `WANDB_RESUME=allow`, so a
collision reads as a request to resume the earlier run.

Retries keep `-aN`. Ids are stored on the record at launch, so only new launches
change. Updates `docs-next/src/content/docs/guides/tools/wandb-integration.md`.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/404" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/run.py
+++ b/modal_training_gym/common/run.py
@@ -438,8 +438,9 @@ def record_resume_checkpoint(
 
 
 def wandb_run_id_for_attempt(training_run_id: str, attempt_count: int) -> str:
-    base_run_id = training_run_id[:8]
-    return base_run_id if attempt_count <= 1 else f"{base_run_id}-a{attempt_count}"
+    return (
+        training_run_id if attempt_count <= 1 else f"{training_run_id}-a{attempt_count}"
+    )
 
 
 def record_wandb_attempt(
--- a/docs-next/src/content/docs/guides/tools/wandb-integration.md
+++ b/docs-next/src/content/docs/guides/tools/wandb-integration.md
@@ -78,14 +78,14 @@ fix — or to drop `wandb=` if you didn't want logging at all.
 
 ## How runs map to W&B runs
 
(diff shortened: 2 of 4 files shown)


modal_training_gym/common/run.py
```diff
@@ -438,8 +438,17 @@ def record_resume_checkpoint(
...
-    base_run_id = training_run_id[:8]
-    return base_run_id if attempt_count <= 1 else f"{base_run_id}-a{attempt_count}"
+    """The tracker run id for an attempt: the training run's own id.
+
+    Exported as ``WANDB_RUN_ID`` with ``WANDB_RESUME=allow``, so it has to be
+    unique per run — a repeated id reads as a request to resume the earlier run
+    rather than to start a new one. A training run id is a random slug plus a
+    digest (``create_hash``) and only the digest is distinguishing, so don't
+    shorten it here.
+    """
```
joyliu-q: Don't need a comment here

joyliu-q (approved):

PR #406, opened 2026-08-16
skip redundant megatron->hf conversions

andrewhinh:
Skip `convert_megatron_checkpoint_to_hf` when HF dir already present.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/406" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/checkpoint.py
+++ b/modal_training_gym/common/checkpoint.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import Enum
-import json
 import os
 import time
 
@@ -20,6 +19,7 @@ from modal_training_gym.deploy_recipes import SglangRecipe, VllmRecipe
 
 
 _CHECKPOINTS_MOUNT_FALLBACK = "/checkpoints"
+_CONVERT_COMPLETE_MARKER = ".training_gym_convert_complete"
 
 
 class CheckpointType(Enum):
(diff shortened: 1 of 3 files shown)


modal_training_gym/common/checkpoint.py
```diff
@@ -173,20 +172,80 @@ def convert_megatron_checkpoint_to_hf(
...
+                    if isinstance(weight_map, dict) and weight_map
+                    else set()
+                )
+                break
+        if shards is None:
+            ready = "model.safetensors" in names or "pytorch_model.bin" in names
+        else:
+            ready = bool(shards) and shards <= names
+    else:
+        ready = False
```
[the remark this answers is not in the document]
andrewhinh: marker is back. volumes background-commit every 5s, so a name-set can look complete while a shard is still truncated. skip only if `.training_gym_convert_complete` is in the listing.

modal_training_gym/common/checkpoint.py
```diff
@@ -173,20 +172,44 @@ def convert_megatron_checkpoint_to_hf(
...
+            path=output_path,
+            timestamp=checkpoint.timestamp,
+            training_run_id=checkpoint.training_run_id,
+            app_name=checkpoint.app_name,
+            checkpoints_volume_name=checkpoints_volume_name,
+            checkpoints_mount_path=checkpoints_mount_path,
+        )
+
+    import modal
+    from modal import App
```
joyliu-q: Why are we reimporting here

joyliu-q: Also why are we importing modal and from modal import App

this is also a local import

andrewhinh: bot slop not caught by slop cop (me)

modal_training_gym/common/checkpoint.py
```diff
@@ -253,10 +274,11 @@ def convert_megatron_to_hf(
         )
         print(f"Converting checkpoint for serving: {cmd}")
         subprocess.run(["bash", "-c", cmd], check=True)
+        with open(os.path.join(output_dir, _CONVERT_COMPLETE_MARKER), "w"):
+            pass
         checkpoints_volume.commit()
```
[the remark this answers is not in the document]
andrewhinh: dismiss: convert already wrote --output-dir (subprocess check=True). live gpu wrote the marker. if the dir were missing, convert would have failed first.

modal_training_gym/common/checkpoint.py
```diff
@@ -76,7 +78,7 @@ def _to_volume_path(checkpoint_dir: str, checkpoints_mount_path: str) -> str:
 
 def _list_checkpoints(train_result: "TrainResult") -> list[Checkpoint]:
     checkpoint_dir = train_result.checkpoint_dir.rstrip("/")
-    if not checkpoint_dir:
+    if checkpoint_dir == "":
```
anli5005, line 81: Not blocking but a lot of what seem to be drive-by changes along these lines (changing truthiness checks to exact comparisons and vice versa)

modal_training_gym/common/checkpoint.py
```diff
@@ -173,20 +173,41 @@ def convert_megatron_checkpoint_to_hf(
...
+    rel = _to_volume_path(output_path, checkpoints_mount_path)
+    try:
+        entries = list(volume.iterdir(rel or "/", recursive=False))
+    except (FileNotFoundError, NotFoundError):
+        entries = []
+    names: set[str] = set()
+    for entry in entries:
+        name = getattr(entry, "path", "").rstrip("/").rsplit("/", 1)[-1]
+        if name:
+            names.add(name)
```
anli5005: I'll defer to you on this one but it is there a way to accomplish this without needing to enumerate the directory?

For example, we could attempt to read the file using `Volume.read_file`, and then if we get a file not found error we proceed with conversion

anli5005 (approved):

modal_training_gym/common/checkpoint.py
andrewhinh: fixed in caf1b9a0. probe the marker with volume.read_file instead of listing the hf dir.

modal_training_gym/common/checkpoint.py
andrewhinh: not blocking. optional string fields use explicit none/empty checks so none and empty stay distinct; leaving the rest.

PR #423, opened 2026-08-19
Render legacy-derived timings in the run timeline

cskitty:
## Summary

Runs whose only timing data lives in the legacy `step_times`/`substep_times` fields showed no Substep Timing section at all, even though the backend already converts them (`legacy_run_to_records`) and returns real per-phase lanes with `metadata.legacy_derived: true`. Cause was a one-line gate in `shouldShowTimingSection`:
 
Requested by: @cskitty


bokeefe123 (approved):
lgtm

PR #432, opened 2026-08-20
Update skills based on codegolf training

rhea24:
from codegolf agent training:

- agent wrongly cancelled a run because it saw reward plateauing but for dapo, reward is based on oversampled and filtered samples, so reward over time is not sufficient to conclude early stop
- often agent starts a monitor with some broken config and because it doesn't fire, the monitor is just broken the whole time -> so be clear in skill about ACTIVELY monitoring the runs, not just passively so (that way, it detects broken configs without wasting as much time)
- add full run final experiment config (e.g. num rollouts/steps) to user confirmation prompt for vague prompts

### DIFF
--- a/skills/agent-driven-training/references/debug-reward.md
+++ b/skills/agent-driven-training/references/debug-reward.md
@@ -17,7 +17,9 @@ DAPO oversamples candidates and filters groups such as all-correct or
 all-incorrect samples. Its reported reward is therefore conditioned on a
 changing retained population and need not rise monotonically. A DAPO plateau
 alone is not a stop signal; compare fixed task metrics, filter/retention counts,
-and representative traces.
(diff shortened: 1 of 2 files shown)


skills/agent-driven-training/references/debug-reward.md
```diff
@@ -76,6 +84,8 @@ false positive or false negative.
   do not launch a full horizon to produce an uninformative curve.
 - Sparse but valid signal: adjust one reward or sampling choice and repeat the
   smoke test.
+- Base model almost never crosses a required correctness gate: propose a
+  stronger model or a justified graded signal/curriculum, then repeat the proof.
```
rhea24:
```suggestion
  stronger model or a justified graded signal/curriculum.
```

joyliu-q (approved):

PR #445, opened 2026-08-22
remove recipe merging

andrewhinh:
Remove `merge_model_recipe` from `TrainConfig` and `TrainingGroup`. Encourages users to use the model recipe classes instead of `SlimeRecipe` or `MilesRecipe`. Not backwards compatible.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/445" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/train.py
+++ b/modal_training_gym/common/train.py
@@ -1,4 +1,4 @@
-import dataclasses as _dc
+import copy
 import os
 import secrets as _secrets
 import sys
@@ -6,7 +6,7 @@ import threading
 import time
 import warnings
 from contextlib import nullcontext
-from typing import Any, TypeVar, cast
+from typing import Any
 
 from pydantic import ConfigDict
 from pydantic.dataclasses import dataclass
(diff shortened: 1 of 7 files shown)


modal_training_gym/common/train.py
```diff
@@ -437,39 +370,41 @@ def _prepare_recipe(self) -> BaseTrainRecipe:
                 "Hugging Face exports are serving artifacts."
             )
         if isinstance(recipe, (MilesRecipe, SlimeRecipe)):
-            recipe = _dc.replace(
+            object.__setattr__(
                 recipe,
-                load=os.path.dirname(self.checkpoint.path.rstrip("/")),
+                "load",
+                os.path.dirname(self.checkpoint.path.rstrip("/")),
```
joyliu-q: Why did we change impl here? It doesn't get rid of merge recipes

andrewhinh: fixed in 2b7f7069

modal_training_gym/common/train.py
```diff
@@ -424,52 +360,49 @@ def _generate_training_run_id(self) -> str:
...
-        )
         if self.checkpoint is None:
-            return recipe
-        if self.checkpoint.checkpoint_type != CheckpointType.megatron:
-            raise TrainingGymConfigError(
-                "Training can only resume from a Megatron checkpoint; "
-                "Hugging Face exports are serving artifacts."
-            )
-        if isinstance(recipe, (MilesRecipe, SlimeRecipe)):
+            recipe = _dc.replace(self.recipe)
```
joyliu-q, line 365: why did we decide to do `_dc.replace` here instead of returning the recipe

andrewhinh: The merge code before this PR already returned a [copy](https://github.com/modal-projects/training-gym/blob/b70265f4d66052d6b74d3e0d2c30ef128509dd0e/modal_training_gym/common/train.py#L426-L444). If we return the recipe directly, Miles (and Slime) would [modify it in-place](https://github.com/modal-projects/training-gym/blob/2b7f7069854e9f3e4d61de29bd969dc84ae8698f/modal_training_gym/frameworks/miles/launcher.py#L474-L580), making `TrainConfig` non-idempotent across calls.

modal_training_gym/common/train.py
```diff
@@ -424,52 +360,49 @@ def _generate_training_run_id(self) -> str:
...
+        _try_validate_model_parallelism(recipe, self.model)
         return recipe
 
     def _build_app(self, training_run_id: str | None = None):
         _warn_if_external_build_app()
         if training_run_id is None:
             training_run_id = self._generate_training_run_id()
-        recipe_type = self.recipe.recipe_type
+        recipe = self._prepare_recipe()
+        recipe_type = recipe.recipe_type
```
anli5005: Is there a reason we still have `recipe_type` instead of just deriving it from the class?

andrewhinh: nope, though perhaps in a follow-up? as this is existing behavior

anli5005 (approved):

PR #481, opened 2026-09-01
remove recipe_type field from TrainRecipe

andrewhinh:
Removes `recipe_type` from train recipes and instead infers from their class.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/481" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/train.py
+++ b/modal_training_gym/common/train.py
@@ -6,15 +6,17 @@ import threading
 import time
 import warnings
 from contextlib import nullcontext
-from typing import Any
-from typing import TypeVar
-from typing import cast
+from typing import Any, TypeVar, cast
+
+from pydantic import ConfigDict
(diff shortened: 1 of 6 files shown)


modal_training_gym/common/train.py
```diff
@@ -397,7 +395,7 @@ class TrainConfig:
     # ── Composed configs (required) ─────────────────────────────────────────
     dataset: DatasetConfig
     model: ModelConfig
-    recipe: BaseTrainRecipe
+    recipe: SlimeRecipe | MilesRecipe
```
anli5005, line 398: Maybe we should make this a type alias of some sort? Just to make it easier when we add other recipe types

anli5005 (approved):

PR #483, opened 2026-09-01
Rework dataset API (v2)

anli5005:
A re-implementation of #377. To recap the changes in that PR:

* Evaluation datasets are now fully separate DatasetConfig instances
* Datasets implement `rows()`, with materialization to disk handled (in most cases) internally
* Dataset initializers no longer take arbitrary keyword arguments

Notable changes *from* #377:

* `DatasetConfig` members that were fields or `@property` properties are now simply methods for better clarity.
* `HuggingFaceDataset` and `HarborDataset` are now primarily used directly via their initializers instead of subclassing.
* Datasets are now materialized to a path based on their `cache_key`, or a random UUID if no cache key is specified.
* `always_prepare` on `DatasetConfig` has been removed entirely. Returning `None` from `cache_key` has the same effect. `HuggingFaceDataset` and `HarborDataset` support an `always_download` argument to opt in to this behavior on the subclass level.
* `HuggingFaceDataset` no longer ignores `apply_chat_template` depending on the values of `input_column` or `output_column`. When `apply_chat_template` is False, `input_column` and `output_column` are used as `input_key` and `label_key`.

Before these changes:

```python
class MyCustomDataset(DatasetConfig):
    input_key = "messages"
    label_key = "label"
    apply_chat_template = True

    def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
        os.makedirs(os.path.dirname(path), exist_ok=True)
        rows = [
            {
                self.input_key: [{"role": "user", "content": prompt}],
                self.label_key: label,
            }
            for prompt, label in prompts
        ]
        Dataset.from_list(rows).to_parquet(path)
        if eval_paths:
            for eval_path in eval_paths.values():
                os.makedirs(os.path.dirname(eval_path), exist_ok=True)
                Dataset.from_list(rows).to_parquet(eval_path)
```

After:

```python
class MyCustomDataset(DatasetConfig):
    def input_key(self) -> str:
        return "messages"

    def label_key(self) -> str:
        return "label"

    def rows(self):
        for prompt, label in prompts:
            yield {
                self.input_key(): [{"role": "user", "content": prompt}],
                self.label_key(): label,
            }
```

Note that while the BFCL dataset was adapted to support the cross-tokenizer distillation tutorial, the Toolathlon dataset has not been migrated for now and is out of scope of this PR.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/483" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


modal_training_gym/common/dataset.py
```diff
@@ -139,112 +106,129 @@ def validate_prepared(self, path: str) -> None:
...
+            self.label_key(): str(row[self.output_column]),
+        }
 
     def to_pandas(self, *, formatted: bool = False):
-        ds = self.load()
-        if formatted:
-            ds = self._format_for_training(ds)
-        return ds.to_pandas()
+        ds = self._load_hf_dataset()
+        return ds.to_pandas(formatted=formatted)
```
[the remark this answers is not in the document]
anli5005: `to_pandas` was removed entirely (tutorial notebooks, the one call site, no longer exist), so this is no longer relevant.

modal_training_gym/common/dataset.py
```diff
@@ -139,112 +106,129 @@ def validate_prepared(self, path: str) -> None:
...
+                "hf_split": self.hf_split,
+                "hf_config": self.hf_config,
+                "input_column": self.input_column,
+                "output_column": self.output_column,
+                "apply_chat_template": self.apply_chat_template,
+                "system_prompt": self.system_prompt,
+                "prompt_template": self.prompt_template,
+                "output_format": self.output_format(),
+            }
+        )
```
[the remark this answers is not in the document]
anli5005: This is existing behavior in the current dataset implementation. In a future PR, we can pin `HuggingFaceDataset` instances to a specific version and then incorporate that into the cache key.

modal_training_gym/common/dataset.py
```diff
@@ -139,112 +106,129 @@ def validate_prepared(self, path: str) -> None:
                 return
         except Exception as e:  # don't shadow the user's real bug with a sniff bug
             print(
-                f"[{type(self).__name__}.validate_prepared] could not sniff "
+                f"[{type(self).__name__}.validate_written] could not sniff "
                 f"{path!r} ({e!r}); skipping schema check."
             )
             return
```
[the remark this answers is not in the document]
anli5005: This is existing behavior and appears to be intentional. Might be worth changing but not in this PR

modal_training_gym/common/dataset.py
```diff
@@ -139,112 +106,125 @@ def validate_prepared(self, path: str) -> None:
...
-        return ds.to_pandas()
+    def _to_chat(self, row):
+        messages: list[dict[str, str]] = []
+        if self.system_prompt:
+            messages.append({"role": "system", "content": self.system_prompt})
+        user_content = self.prompt_template.format(input=row[self.input_column])
+        messages.append({"role": "user", "content": user_content})
+        return {
+            self.input_key(): messages,
+            self.label_key(): str(row[self.output_column]),
```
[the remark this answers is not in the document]
anli5005: Fixing by setting `apply_chat_template` to False, which disables this formatting.

Note that this differs slightly from `HuggingFaceDataset`'s previous behavior, where it would silently ignore `apply_chat_template` depending on the values of `input_column` and `output_column`. This no longer exists, and instead the column names are used as the input/label keys when `apply_chat_template` is False.

modal_training_gym/common/launcher_helpers.py
```diff
@@ -209,25 +209,30 @@ def run_download_phase(
...
 
+def write_dataset_if_needed(dataset: Any, path: str) -> bool:
+    """Write and validate a dataset unless its cached materialization exists."""
+    if os.path.exists(path):
+        dataset.validate_written(path)
+        return False
+    os.makedirs(os.path.dirname(path), exist_ok=True)
+    print(f"Writing dataset ({path})...")
+    dataset.write(path)
+    dataset.validate_written(path)
```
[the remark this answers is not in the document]
anli5005: Existing issue; I can imagine using something like a Modal dict/queue but probably in scope for a separate PR

modal_training_gym/train_recipes/base.py
```diff
@@ -97,31 +110,50 @@ def validate_model_parallelism(self, model: "ModelConfig") -> None:
...
+    ) -> None:
+        if eval_ds is None:
+            return
+        for key_method in ("input_key", "label_key"):
+            train_key = getattr(ds, key_method)()
+            eval_key = getattr(eval_ds, key_method)()
+            if train_key != eval_key:
+                raise TrainingGymConfigError(
+                    f"Training and evaluation datasets must use the same "
+                    f"{key_method}(): got {train_key!r} and {eval_key!r}."
```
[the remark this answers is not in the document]
anli5005: Multimodal datasets are subject to change, and setting custom modalities is not common to all datasets. This is enough of an edge case that I would rather handle this in another PR if possible

modal_training_gym/common/launcher_helpers.py
```diff
@@ -209,25 +209,30 @@ def run_download_phase(
         flush_status_reporter(timeout_seconds=2.0)
 
 
+def write_dataset_if_needed(dataset: Any, path: str) -> bool:
+    """Write and validate a dataset unless its cached materialization exists."""
+    if os.path.exists(path):
+        dataset.validate_written(path)
+        return False
```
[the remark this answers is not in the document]
anli5005: This is an existing issue and should be addressed in a future PR, given potential the performance implications here

modal_training_gym/common/dataset.py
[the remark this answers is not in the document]
anli5005: This is equivalent to the existing `validate_prepare` behavior

docs-next/src/content/docs/guides/start/dataset.md
```diff
@@ -67,68 +73,58 @@ A [Harbor dataset](https://gym.modal.dev/reference/harbordataset) provides a ser
 ```python
 from modal_training_gym import HarborDataset
 
-
-class HelloWorld(HarborDataset):
-    dataset_name="harbor/hello-world"
+hello_world = HarborDataset(dataset_name="harbor/hello-world")
```
andrewhinh: nit: `hello_dataset`

docs-next/src/content/docs/guides/start/dataset.md
```diff
@@ -67,68 +73,58 @@ A [Harbor dataset](https://gym.modal.dev/reference/harbordataset) provides a ser
...
-    label_key = "label"
-    apply_chat_template = True
-
-    def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
-        os.makedirs(os.path.dirname(path), exist_ok=True)
-        rows = [
-            {
-                self.input_key: [{"role": "user", "content": prompt}],
-                self.label_key: label,
+    def input_key(self) -> str:
```
andrewhinh, line 118: as a user, it's not immediately clear why `input_key` and `label_key` have to be methods. maybe a sentence or two above would be very helpful.

docs-next/src/content/docs/guides/tools/agent.md
```diff
@@ -56,28 +56,15 @@ SYSTEM_PROMPT = (
...
-        from datasets import load_dataset
-
-        ds = load_dataset(self.hf_repo, self.hf_config, split=self.hf_split)
-        ds = ds.filter(
-            lambda r: not r["input"].strip() and 60 <= len(r["output"]) <= 600
-        )
-        if self.n_rows:
-            ds = ds.select(range(min(self.n_rows, len(ds))))
-        return ds
+def rhyme_dataset(n_rows: int) -> HuggingFaceDataset:
```
andrewhinh: wdyt about inlining this?

modal_training_gym/common/dataset.py
```diff
@@ -546,117 +576,92 @@ def load(self, split: Literal["all", "train", "eval"] = "all") -> Any:
...
+    def input_key(self) -> str:
+        return "prompt"
+
+    def label_key(self) -> str:
+        return "label"
 
-    @property
-    def rows(self) -> list[DatasetRow]:
-        return self._rows
+    def source_rows(self) -> Iterable[dict[str, Any]]:
```
andrewhinh, line 652: why is it named `source_rows`?

scripts/validate_model_configs.py
```diff
@@ -386,6 +386,7 @@ def run_base_training(
     train_config = TrainConfig(
         model=model_config,
         dataset=dataset,
+        eval_dataset=dataset,
```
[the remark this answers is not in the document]
andrewhinh: besides Devin's point, do we need to pass this?

tutorials/audio_asr.py
```diff
@@ -69,17 +69,19 @@ def score_transcript(response: str, label: str) -> float:
 # resolve them in a custom `generate` function.
 
 class LibriSpeechASRDataset(MultimodalDataset):
-    modality = "audio"
     hf_repo = "hf-internal-testing/librispeech_asr_dummy"
     hf_config = "clean"
```
andrewhinh, line 73: why can you avoid specifying `input_format` here?

andrewhinh (approved):
added questions for clarification

tutorials/audio_asr.py
anli5005: `input_format` only exists on `HuggingFaceDataset`. In this case since you're directly subclassing `MultimodalDataset` you instead override `apply_chat_template` yourself.

Can change this if it's unclear

modal_training_gym/common/dataset.py
anli5005: `rows` applies some minor formatting to each source row. Namely, it takes any media stored under the `media` key, encapsulates it in a list, and then writes it under the correct key. You can't directly override `rows`, because then there would be no place to apply this formatting.

Can think of a better name or interface, but also we could just remove the formatting entirely. In general I want to rework `MultimodalDataset` as a whole since the API surface is not very clear, but was saving that for another PR down the road

scripts/validate_model_configs.py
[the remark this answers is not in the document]
anli5005: We need to pass *something* here otherwise the validation CI breaks with the default `eval_interval` ([example](https://github.com/modal-projects/training-gym/actions/runs/33550907757/job/99999941484)). We could set `eval_interval` to None or pick a slightly different eval dataset though.

docs-next/src/content/docs/guides/tools/agent.md
anli5005: I'll make it a plain variable so we don't disrupt the flow of the guide

docs-next/src/content/docs/guides/start/dataset.md
anli5005: I think this would be helpful to include in a migration guide, but might be best to just leave it as "part of the DatasetConfig interface" here.

The actual technical reason is that subclasses might want to return a value here that depends on other configurable properties, and making these fields would open the door for clients of those subclasses to mess with that. But if we're not explaining what `input_key` and `label_key` actually do in this guide, this is probably best left as an implementation detail for users who are just starting out. Happy to consider otherwise though

andrewhinh: ok fine by me

docs-next/src/content/docs/guides/tools/agent.md
andrewhinh: ok sg

modal_training_gym/common/dataset.py
andrewhinh: ok sg

tutorials/audio_asr.py
andrewhinh: perhaps something in the dataset guide that clarifies the differences between the two would be helpful

modal_training_gym/common/dataset.py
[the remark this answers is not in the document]
anli5005: HuggingFaceDataset PR: https://github.com/modal-projects/training-gym/pull/529

PR #487, opened 2026-09-01
Terminalize launch() runs whose Modal app never starts

joyliu-q:
Context: omkaar-dev's dashboard showed ~57 `Pending` runs created ~20s apart between 23:12–23:25 UTC, all `status=running`, `modal_app_id=""`, `framework_status=initializing`. They come from `TrainConfig.launch()`: the `TrainingRun` record is saved *before* `_build_app()` / `app.run(detach=True)`, so if either raises, the record is left in `running` with no app id. The reconciler only catches these via `stale_pre_active_no_modal_app` after `PRE_APP_TIMEOUT_SECONDS` (90 min).

Fix: `launch()` now wraps the app start and, if it fails before an app id was recorded, terminalizes the record immediately:

```
try:
    function_call = self._start_detached_app(run_record, ...)   # _build_app + app.run + spawn (moved, unchanged)
except BaseException as exc:
    if not run_record.modal_app_id:
        _terminalize_unlaunched_run(run_record, exc)   # FAILED (+error_message) / STOPPED on KeyboardInterrupt
    raise
```

`terminal_reason` is `launch_failed_before_modal_app` / `launch_interrupted`. Once the app id is recorded the existing reconciler path (`stale_modal_app_terminated`) is the right owner, so nothing changes there. Save failures are logged and swallowed, as with the other best-effort saves in `launch()`.

Not changed by this PR (findings from the omkaar-dev investigation): the cron reconciler itself works — it terminalized 6 app-backed orphans at 23:30 UTC, right after the dashboard app was redeployed at 23:11 (`modal app history` shows only v1 from 23:11, so the previous dashboard app — and its cron — was gone between ~20:40 and 23:11, which is why runs cancelled earlier in the day sat as Pending for hours). Stopping the dashboard app in the Modal UI kills the cleanup loop until the next `launch()` redeploys it.

## Checklist

N/A — library change, not an example.

Link to Devin session: https://modal.devinenterprise.com/sessions/14667a3b70a94561a404a7604b9da11a
Open in Devin Desktop: https://modal.devinenterprise.com/desktop/session/14667a3b70a94561a404a7604b9da11a?variant=devin
Requested by: @joyliu-q

### DIFF
--- a/modal_training_gym/common/train.py
+++ b/modal_training_gym/common/train.py
@@ -18,7 +18,11 @@ from modal_training_gym.common.framework import Framework
 from modal_training_gym.common.ids import create_hash
 from modal_training_gym.common.modal_urls import modal_app_dashboard_url
 from modal_training_gym.common.models import ModelConfig
-from modal_training_gym.common.run import TrainingRun, metric_run_id_for_attempt
+from modal_training_gym.common.run import (
+    TrainingRun,
+    TrainingRunStatus,
+    metric_run_id_for_attempt,
+)
 from modal_training_gym.common.status import (
     FrameworkStatus,
     MilesStatus,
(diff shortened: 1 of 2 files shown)


anli5005:
Closing in favor of #502

PR #489, opened 2026-09-02
Page the dashboard's training run list instead of shipping every run to the phone

joyliu-q:
The dashboard freezes/crashes mobile Safari because `/api/runs` returns the entire history as one array — 381 runs, 1.49 MB of JSON — every 5s (including while the tab is hidden), and the table mounts a row for each of them (~20k DOM nodes).

The list endpoint now filters, searches, sorts and pages server-side, and drops the fields the table never renders:

```python
# /api/runs?limit=100&offset=0&q=…&status=failed&recipe=slime
filtered = filter_run_summaries(summaries, facets=…, query=q, offset=offset, limit=limit, sort_by="created")
return JSONResponse([s.model_dump(mode="json", exclude={"config", "step_times", "substep_times"}) for s in filtered])
```

`config` alone was ~60% of the payload; the drawer/detail page fetches the full record from `/api/runs/{id}` on open, so nothing is lost. Sorting is by `created_at` rather than `updated_at` so pages don't reshuffle under the client when a run reports progress.

Paging breaks every total on the page (they were counted client-side from the full array), so `/api/runs/counts` returns just the numbers:

```json
{"total": 381, "matching": 94, "status": {"failed": 219, …}, "recipe": {"slime": 269, …}, "group": {…}}
```

Bucket counts cover every run (that's what the chips would select); `matching` is how many rows paging can still reach. The frontend keeps only the loaded window in `runs`, asks for the next 100 when you scroll near the end, and restarts at page one (scrolled back to the top) when the search or chips change.

Also in this branch, from the same investigation: a single shared clock for all `TimeAgo` components instead of one `setInterval` each, polling paused while `document.hidden`, and no overlapping refreshes (a slow response used to stack whole copies of the payload in memory).

Measured against real data (381 runs) in a 390×844 mobile viewport: first paint fetches 250 KB instead of 1.49 MB, 5.4k DOM nodes instead of 19.7k; the earlier heap/main-thread numbers on this branch were 56–84 MB → 24–31 MB and 7.0s → 1.3s of long tasks under 6× CPU throttling. That's desktop emulation — I haven't been able to reproduce on a real iPhone, so worth a check on your phone after it deploys.

## Checklist

Not an example; the example checklist doesn't apply.

Link to Devin session: https://modal.devinenterprise.com/sessions/95e5ad2ac22742e98d76804b1c04343b
Open in Devin Desktop: https://modal.devinenterprise.com/desktop/session/95e5ad2ac22742e98d76804b1c04343b?variant=devin
Requested by: @joyliu-q

### DIFF
--- a/dashboards/frontend/src/App.svelte
+++ b/dashboards/frontend/src/App.svelte
@@ -7,13 +7,31 @@
   import TrainingPage from "./pages/TrainingPage.svelte";
   import TrainingRunDetailPage from "./pages/TrainingRunDetailPage.svelte";
   import EvalsPage from "./pages/EvalsPage.svelte";
-  import { fetchRuns, fetchEvals, fetchEvalDetail } from "./lib/api.js";
+  import {
+    fetchRuns,
+    fetchRunCounts,
+    fetchEvals,
+    fetchEvalDetail,
+  } from "./lib/api.js";
   import logoSvg from "./lib/logo.svg";
   import { fmtDuration } from "./lib/format.js";
 
   const DOCS_URL = "https://gym.modal.dev";
 
-  let allRuns = $state([]);
+  // The server filters, sorts and pages the run list: `runs` only ever holds
+  // the rows the page has asked for. Every total on the page comes from
+  // `runCounts` instead — the whole history is far too much JSON for a phone to
+  // download and keep resident just to count it.
+  const RUNS_PAGE_SIZE = 100;
+  let runs = $state([]);
+  let loadedRunCount = $state(RUNS_PAGE_SIZE);
+  let runCounts = $state({
+    total: 0,
+    matching: 0,
+    status: {},
+    recipe: {},
+    group: {},
+  });
   let allEvals = $state([]);
   let loading = $state(true);
   let loadingEvals = $state(false);
@@ -117,10 +135,8 @@
     };
   });
 
-  function getRecipe(run) {
-    return run.recipe || run.framework || "(untagged)";
-  }
-
+  // Mirrors the server's facet buckets (`run_facet_values`), which is what the
+  // filter chips and counts are keyed by.
   const NO_GROUP = "(no group)";
 
   function getGroup(run) {
@@ -153,10 +169,6 @@
     return value != null ? String(value) : "";
   }
 
-  function includesText(value, query) {
-    return safeText(value).toLowerCase().includes(query);
-  }
-
   function getErrorMessage(value) {
     if (value instanceof Error) return value.message;
     if (typeof value === "string") return value;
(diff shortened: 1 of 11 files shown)


dashboards/frontend/src/lib/api.js
```diff
@@ -12,13 +12,56 @@ async function getErrorFromResponse(res) {
...
+  const body = await res.text();
+  try {
+    return JSON.parse(body);
+  } catch {
+    const contentType = res.headers.get("content-type") || "unknown type";
+    throw new Error(
+      `${what}: expected JSON, got ${contentType} (HTTP ${res.status}, ${body.length} bytes)`,
+    );
+  }
+}
```
anli5005, line 30: This (and the later `api_not_found` route) seem like a drive-by change - was there anything specific you were running into here?

modal_training_gym/_dashboard.py
```diff
@@ -840,33 +842,92 @@ async def _get_run_or_404(training_run_id: str) -> TrainingRun:
...
+
+    # ``response_model`` is left off: FastAPI ignores ``response_model_exclude``
+    # for sequence response models, so the exclusion is applied here instead.
+    @web.get("/api/runs")
     async def runs(
         request: Request,
         since: int | None = None,
         limit: int | None = None,
+        offset: int = 0,
+        q: str = "",
```
anli5005, line 890: Surely there's a better way to consume three fixed query parameters (`status`, `recipe`, and `group`) than passing a request object to a `_requested_facets` function and expecting readers of this code to just know about it

dashboards/frontend/src/pages/TrainingPage.svelte
```diff
@@ -171,12 +198,53 @@
...
+    Math.max(matchingRunCount - filteredRuns.length, 0),
+  );
+
+  function growOnScroll(event) {
+    if (!hasMoreRuns) return;
+    const el = event.currentTarget;
+    if (el.scrollTop + el.clientHeight >= el.scrollHeight - 600) {
+      onLoadMore();
+    }
+  }
```
anli5005: Why do we still have this if we also use `IntersectionObserver`? [Browser support for the latter seems widespread](https://caniuse.com/?search=IntersectionObserver) so if the reason is just compatibility we should get rid of this separate event handler

dashboards/frontend/src/pages/TrainingPage.svelte
```diff
@@ -420,6 +492,12 @@
           {/each}
         </div>
       {/if}
+
+      {#if hiddenRunCount > 0}
+        <div bind:this={sentinel} class="page-empty">
+          Showing {filteredRuns.length} of {matchingRunCount} runs
+        </div>
+      {/if}
```
anli5005: I don't see a reason we need to only show this when there are hidden runs

dashboards/frontend/src/App.svelte
```diff
@@ -7,13 +7,31 @@
...
   import logoSvg from "./lib/logo.svg";
   import { fmtDuration } from "./lib/format.js";
 
   const DOCS_URL = "https://gym.modal.dev";
 
-  let allRuns = $state([]);
+  // The server filters, sorts and pages the run list: `runs` only ever holds
+  // the rows the page has asked for. Every total on the page comes from
+  // `runCounts` instead — the whole history is far too much JSON for a phone to
+  // download and keep resident just to count it.
```
anli5005: I'm not sure I like the reasoning in this comment. Modern phones are more than able to handle ~1 MB of JSON. I feel like the actual problem is twofold:

1. We keep refreshing this entire JSON file every 5 seconds, which creates far more network traffic than necessary
2. We show and update more DOM nodes than is actually necessary, hurting performance

Pagination is still a good solution for both, it just helps to be more precise. In this case though this is a clear slop comment (what else would the code be doing other than pagination?) so should probably just get rid of the comment altogether

dashboards/frontend/src/App.svelte
```diff
@@ -94,22 +112,33 @@
...
       if (activePage === "training" && activeTrainingRunId) return;
       void load();
     }, 5000);
 
+    const onVisibilityChange = () => {
+      if (document.hidden) return;
+      if (activePage === "training" && activeTrainingRunId) return;
+      void load();
+    };
+    document.addEventListener("visibilitychange", onVisibilityChange);
```
anli5005: Wouldn't a better solution be to just clear the refresh timer entirely when the document is hidden?

dashboards/frontend/src/App.svelte
```diff
@@ -167,6 +192,76 @@
...
+      group: activeGroups,
+    };
+    const facets = {};
+    for (const [name, selected] of Object.entries(selections)) {
+      const universe = universes[name];
+      if (!universe.length) continue;
+      if (universe.every((value) => selected.has(value))) continue;
+      facets[name] = [...selected];
+    }
+    return facets;
```
anli5005, line 226: can't wait for `Iterator.zipKeyed` to gain widespread support

(don't actually use it here it's pretty much only on firefox at the moment)

anli5005 (approved):

PR #492, opened 2026-09-02
Count num_rollout from zero when continuing from a checkpoint

joyliu-q:
## Summary

`TrainConfig(checkpoint=...)` only set `recipe.load`, so slime/miles read the stored iteration `N` from the checkpoint and ran `range(N + 1, num_rollout)`. A continuation run with `num_rollout <= N + 1` therefore ran zero rollouts and reported itself as completed; users had to keep bumping `num_rollout` past the parent run's step count.

Both frameworks accept `--start-rollout-id`, which overrides the loaded iteration in that loop. This PR:

- adds a `start_rollout_id: int | None = None` field to `SlimeRecipe` and `MilesRecipe` (emitted as `--start-rollout-id`)
- `TrainConfig._prepare_recipe` sets `start_rollout_id=0` when a `checkpoint` is given (unless the recipe set it explicitly), so `num_rollout` is the number of steps *this* run performs and its checkpoints are numbered `iter_0000000...` under its own `save_root`
- the launchers' auto-resume branch (a retry finding a checkpoint in the run's own `save_root`) forces `start_rollout_id=None` for the built command, so crash-retries of a continuation run still pick up at the run's last saved iteration instead of restarting at 0

Behavior for plain crash-retries of a non-continuation run is unchanged (`_validate_resume_checkpoint` still guards the `num_rollout` target there).

Not tested on a live cluster; verified via `tests/test_train_checkpoint.py` (new cases) and inspection of slime's `train.py` / `placement_group.py` and miles' equivalents, which honor a user-set `args.start_rollout_id`.

## Checklist

- [x] N/A — SDK change, not an example

Link to Devin session: https://modal.devinenterprise.com/sessions/a83bfe400bc44aa0b1d8afc5528c56a9
Open in Devin Desktop: https://modal.devinenterprise.com/desktop/session/a83bfe400bc44aa0b1d8afc5528c56a9?variant=devin
Requested by: @joyliu-q

### DIFF
--- a/modal_training_gym/common/launcher_utils.py
+++ b/modal_training_gym/common/launcher_utils.py
@@ -469,6 +469,33 @@ def prepare_launch_config(
             object.__setattr__(cfg, field, path)
 
 
+def drop_materialized_config_key(cfg: Any, key: str) -> None:
+    """Remove ``key`` from the recipe's already-materialized escape-hatch YAML.
+
+    Runs after ``prepare_launch_config`` has replaced the escape-hatch dict
+    with a file path, so the YAML is rewritten and the recorded keys updated
+    so ``_emit_fields`` stops suppressing the same-named flag.
+    """
+    import yaml
+
+    escape_hatch = getattr(cfg, "_ESCAPE_HATCH_FIELD", None)
+    if not escape_hatch:
+        return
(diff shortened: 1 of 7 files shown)


andrewhinh (approved):
lgtm

PR #493, opened 2026-09-02
Add partial rewards

joyliu-q:
## Checklist

## Outside contributors


modal_training_gym/common/sample.py
```diff
@@ -24,6 +24,25 @@ class TraceSpan(BaseModel):
...
 
+class RewardEvent(BaseModel):
+    """One incremental or checkpoint-window reward emitted during a sample.
+
+    ``reward`` is the credit assigned to this event/window, not the final
+    sample score. Token offsets are relative to the sample response/loss-mask
+    sequence, when the environment can provide them. Keeping this separate
+    from ``Sample.score`` lets dashboards inspect shaped rewards without
+    changing the scalar reward contract used by current trainers.
+    """
```
joyliu-q, line 37: Add a comment here that it is experimental

PR #503, opened 2026-09-04
Partition SWE-bench-style datasets for the agentic recipe

zhouhelena1:
Adds SWE-rebench-V2 partition tooling for agentic Slime training.

- `prepare` converts tasks and creates repository-disjoint train/eval splits, targeting 20% eval, plus nested subsets. Failed refreshes preserve the existing dataset.
- `mixed` selects tasks with both successful and failed episodes from an eval dump.

The pinned converter supports Python only. The actual eval percentage is reported; the full-dataset ratio is not yet validated.


scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,689 @@
...
+            ):
+                print(
+                    f"[harbor] unpacked {completed}/{len(bundles)} archives "
+                    f"({task_count} tasks, {time.monotonic() - started:.0f}s)"
+                )
+
+    def convert(self, root: Path) -> list[dict[str, Any]]:
+        converted_path = root / "all.converted.jsonl"
+        if converted_path.is_file() and converted_path.stat().st_size:
+            return read_jsonl(converted_path)
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. partition() now records the inputs the conversion depends on (repo, resolved Hub revision, dataset key, metadata namespace, fork translator revision, archive list) in all.converted.json and reuses all.converted.jsonl only when that record matches; on any change it drops the extracted task directories and markers and rebuilds from the archives.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,689 @@
...
+        volumes["/checkpoints"] = modal.Volume.from_name(
+            args.checkpoints_volume, create_if_missing=False
+        )
+    remote_options: dict[str, Any] = {
+        "image": _image(training_recipe),
+        "volumes": volumes,
+        "timeout": 24 * 60 * 60,
+    }
+    if args.command == "prepare":
+        remote_options["secrets"] = [modal.Secret.from_name("huggingface-secret")]
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. prepare now attaches hf_secrets() from modal_training_gym.common, which forwards a local HF_TOKEN, falls back to the named huggingface-secret if it exists, and otherwise runs unauthenticated.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,690 @@
...
+            task_roots = {
+                member.filename.split("/", 1)[0]
+                for member in members
+                if "/" in member.filename
+            }
+            if task_roots and all(
+                (tasks_root / task_root / "task.toml").is_file()
+                for task_root in task_roots
+            ):
+                return len(task_roots)
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. Each archive is extracted into a staging directory, its task directories are moved into place replacing any partial ones, and a per-archive .extracted marker is written only after that completes; retries redo any archive without a marker.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,690 @@
...
+        if line.strip()
+    ]
+
+
+def write_jsonl(path: Path, rows: list[dict[str, Any]]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(
+        "".join(json.dumps(row, ensure_ascii=False) + "\n" for row in rows),
+        encoding="utf-8",
+    )
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. All outputs (subsets, provenance sidecars, the conversion record) go through write_text(), which writes to a sibling temp file and os.replace()s it into place.

scripts/partition_swe_dataset.py
```diff
@@ -0,0 +1,690 @@
...
+    mixed.add_argument("--checkpoints-volume", required=True)
+    mixed.add_argument("--replace", action="store_true")
+    args = parser.parse_args()
+
+    dataset_root = args.dataset_root
+    if args.command == "prepare":
+        dataset_root = dataset_root or args.hf_repo.replace("/", "_")
+    if not dataset_root:
+        parser.error("--dataset-root is required for mixed")
+    root = f"{DATA_PATH}/{dataset_root}"
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. dataset_root_name() rejects empty, ., .., and anything containing a path separator, and main() applies it to both --dataset-root and the default derived from --hf-repo.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,690 @@
...
+    checkpoint: str | None,
+    replace: bool,
+    volume_name: str,
+):
+    import torch
+
+    dump = Path(probe_dump)
+    if not dump.is_file():
+        raise FileNotFoundError(f"probe dump does not exist: {dump}")
+    payload = torch.load(dump, weights_only=False)
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. The dump is loaded with weights_only=True; the fork's save_debug_rollout_data writes sample.to_dict() payloads of plain dicts, lists, and scalars, so nothing legitimate needs the unrestricted unpickler.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,690 @@
...
+                for task_root in task_roots
+            ):
+                return len(task_roots)
+            for member in members:
+                target = (tasks_root / member.filename).resolve()
+                if not target.is_relative_to(destination):
+                    raise ValueError(
+                        f"archive {bundle.name} contains unsafe path {member.filename!r}"
+                    )
+            archive.extractall(tasks_root)
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in 611e1db6. safe_extract rejects any member whose external attributes mark it as a symlink before extracting, and extraction now targets a staging directory so a rejected archive leaves nothing behind.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+from modal_training_gym.frameworks.slime.launcher import (
+    SLIME_IMAGE,
+    _slime_git_overlay_command,
+)
+from modal_training_gym.train_recipes.base import DATA_PATH
+from modal_training_gym.train_recipes.slime_recipe import (
+    Qwen3_6_27B_Recipe_Agentic,
+)
+
+TRAIN_SIZES = (100, 300, 1000)
```
shariqm-modal: TRAIN_SPLIT_SIZES

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+    for position in range(1, count + 1):
+        available = [
+            language
+            for language, indices in by_language.items()
+            if selected_counts[language] < len(indices)
+        ]
+        language = max(
+            available,
+            key=lambda value: (
+                position * len(by_language[value]) / len(rows) - selected_counts[value],
```
shariqm-modal: Can we replace this lambda with a separate non-lambda function? It's not clear what it's doing. It looks like there's some distribution of languages in the full dataset and we're trying select `count` rows from `rows` (args def func def) such that it matters the full dataset distribution. Would be good to make this clear.

scripts/partition_swe_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+def _group(row: dict[str, Any], suffix_pattern: str) -> str:
+    task_path = str((row.get("metadata") or {}).get("task_path") or "")
+    if not task_path:
+        raise ValueError("converted row is missing metadata.task_path")
+    return re.sub(suffix_pattern, "", Path(task_path).name)
+
+
+def nested_subset(
+    rows: list[dict[str, Any]],
+    count: int,
```
shariqm-modal, line 138: can you make it clear what these args mean with a comment or sth else

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+    value = (row.get("metadata") or {}).get(namespace) or {}
+    return value if isinstance(value, dict) else {}
+
+
+def _language(row: dict[str, Any], namespace: str) -> str:
+    metadata = _metadata(row, namespace)
+    return str(metadata.get("language_bucket") or metadata.get("language") or "?")
+
+
+def _group(row: dict[str, Any], suffix_pattern: str) -> str:
```
shariqm-modal: could we call this function task_group?

And add a comment that says something like, "extract task_group, often a repo name. e.g. aws_aws-cli__2819 maps to aws/aws-cli

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+    SLIME_IMAGE,
+    _slime_git_overlay_command,
+)
+from modal_training_gym.train_recipes.base import DATA_PATH
+from modal_training_gym.train_recipes.slime_recipe import (
+    Qwen3_6_27B_Recipe_Agentic,
+)
+
+TRAIN_SIZES = (100, 300, 1000)
+EVAL_FRACTION = 0.2
```
shariqm-modal: EVAL_SPLIT_FRACTION

Add can you put it above the TRAIN_SPLIT_SIZES and make it clear they are made from the remaining 0.8 fraction

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+
+def repo_disjoint_split(
+    rows: list[dict[str, Any]],
+    *,
+    eval_fraction: float,
+    seed: int,
+    metadata_namespace: str,
+    group_suffix_pattern: str = r"__\d+$",
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+    """Create an approximately language-balanced, repository-disjoint split."""
```
shariqm-modal: Can we update to something more descriptive, can riff on this:

"""Create an approximately language-balanced, repository-disjoint split.

This function first splits off an eval fraction section. It's first priority is to make the task_groups disjoint so that evals do not contain any task_groups from train. In practice task_groups are github repos and we don't want overlap across train & eval for a single repo.

The above has to be balanced with making sure languages are evenly distributed across train/eval, this is the second priority but still important.
"""

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+        for group, indices in groups.items()
+    }
+    language_totals = Counter(_language(row, metadata_namespace) for row in rows)
+    language_groups = {
+        language: {
+            group for group, counts in group_languages.items() if counts[language]
+        }
+        for language in language_totals
+    }
+    target_total = round(len(rows) * eval_fraction)
```
shariqm-modal: rename to `eval_target_num_rows`

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+    language_targets = {
```
shariqm-modal: rename to `eval_target_language_counts`

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+            for language in group_languages[group]
+        )
+
+    def add(group: str) -> None:
+        nonlocal selected_total
+        selected.add(group)
+        selected_total += len(groups[group])
+        eval_languages.update(group_languages[group])
+
+    for language in sorted(
```
shariqm-modal: Add comment for this `for` loop, e.g.: Ensure the eval set has every language represented by preselecting at least one repo for each language

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+        for language in language_totals
+    }
+    target_total = round(len(rows) * eval_fraction)
+    language_targets = {
+        language: count * eval_fraction for language, count in language_totals.items()
+    }
+
+    rng = random.Random(seed)
+    tie_break = {group: rng.random() for group in sorted(groups)}
+    selected: set[str] = set()
```
shariqm-modal: rename to `eval_selected`

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+    selected_total = 0
+    eval_languages: Counter[str] = Counter()
+
+    def can_select(group: str) -> bool:
```
shariqm-modal: rename to `is_needed_for_eval`

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+        selected.add(group)
+        selected_total += len(groups[group])
+        eval_languages.update(group_languages[group])
+
+    for language in sorted(
+        language_totals, key=lambda value: (language_totals[value], value)
+    ):
+        if len(language_groups[language]) < 2 or eval_languages[language]:
+            continue
+        candidates = [group for group in language_groups[language] if can_select(group)]
```
shariqm-modal: rename to `eval_candidates`

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,768 @@
...
+        if candidates:
+            add(
```
shariqm-modal: add coment: "pick the task_group with the least amount of tasks"

shariqm-modal (commented):
got halfway through, will do another pass later.

It would be good to make this code readable for a human as dataset splits are very important to do right and so that quality papers can be written about it. Maybe you can extrapolate my thoughts to the rest of the file.

scripts/partition_swe_dataset.py
zhouhelena1: it's in the docstring below

scripts/partition_harbor_dataset.py
zhouhelena1: I renamed to `can_move_to_eval` instead since it's checking whether the task group / repo can move from train to eval rather than it being needed for eval

scripts/partition_swe_dataset.py
```diff
@@ -0,0 +1,899 @@
...
+        instance_id: {
+            "solved": counts[0],
+            "gradeable": counts[1],
+            "total": counts[2],
+        }
+        for instance_id, counts in sorted(totals.items())
+    }
+    if any(counts["total"] > n_samples for counts in result.values()):
+        raise ValueError(f"probe contains more than {n_samples} episodes for a task")
+    return result
```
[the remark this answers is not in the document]
zhouhelena1: fixed this, so mixed now requires every task to have --n-samples episodes to match eval shape

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,922 @@
...
+        key=lambda instance_id: (
+            abs(2 * counts[instance_id]["solved"] - n_samples),
+            instance_id,
+        ),
+    )
+    missing = [
+        instance_id for instance_id in selected_ids if instance_id not in indexed
+    ]
+    if missing:
+        raise ValueError(f"probe references {len(missing)} tasks absent from {source}")
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in ec43d36d. write_mixed_subset now compares the full set of probe instance ids against the source split before filtering and raises with counts of source tasks missing from the probe and probe tasks not in the source; tests cover truncated, wrong-split, and both.

scripts/partition_swe_dataset.py
```diff
@@ -0,0 +1,922 @@
...
+        rows = self.cached_rows(root, source)
+        if rows is None:
+            self.clear_extracted(root)
+            self.unpack(root)
+            rows = self.convert(root)
+            write_text(
+                root / "all.converted.json",
+                json.dumps(source, indent=2, sort_keys=True) + "\n",
+            )
+        return write_partitions(root, rows, metadata_namespace=self.metadata_namespace)
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in ec43d36d. clear_extracted, which partition() runs whenever the conversion source changes, now also deletes every *-mixed-reward-* JSONL and sidecar under the dataset root, so derived subsets cannot outlive the rows they were filtered from.

scripts/partition_harbor_dataset.py
```diff
@@ -0,0 +1,922 @@
...
+    def source_record(self, root: Path, revision: str) -> dict[str, Any]:
+        """Every input the converted rows depend on; a change invalidates them."""
+        return {
+            "hf_repo": self.hf_repo,
+            "revision": revision,
+            "dataset_key": self.dataset_key,
+            "metadata_namespace": self.metadata_namespace,
+            "translator_revision": self.translator_revision,
+            "archives": [bundle.name for bundle in self.bundles(root)],
+        }
```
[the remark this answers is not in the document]
zhouhelena1: Fixed in ec43d36d. source_record now maps each archive name to its SHA-256 and records the tasks.csv hash, so replaced bytes under an unchanged revision invalidate the cached conversion.

scripts/partition_swe_dataset.py
```diff
@@ -0,0 +1,867 @@
...
+        "probe_dump": probe_dump,
+        "selected_instance_ids": selected_ids,
+        "sha256": sha256(output_path),
+    }
+    write_text(metadata_path, json.dumps(provenance, indent=2, sort_keys=True) + "\n")
+    return output_path, provenance
+
+
+class SweBenchSource:
+    """Stream a SWE-bench-style dataset, render each row as a Harbor task, and convert it."""
```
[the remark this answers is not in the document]
zhouhelena1: I updated the pr desc

scripts/partition_swe_dataset.py
```diff
@@ -0,0 +1,872 @@
...
+    _slime_git_overlay_command,
+)
+from modal_training_gym.train_recipes.base import DATA_PATH
+from modal_training_gym.train_recipes.slime_recipe import (
+    Qwen3_6_27B_Recipe_Agentic,
+)
+
+# Fraction of tasks that go to eval. Remainder go to train.
+EVAL_SPLIT_FRACTION = 0.2
+EVAL_SPLIT_SIZES = (4, 100, 300)
```
shariqm-modal, line 66: do we need different sizes?

shariqm-modal (approved):
lgtm overall! We can refine more later.

PR #505, opened 2026-09-04
Add the coding agent tutorial

zhouhelena1:
Stacked on #557 (`helena/multiturn-ui`), which contains the dashboard rendering changes.

Adds a tutorial for training Qwen3.6-27B coding agents on SWE-rebench V2 through multi-turn interaction with Harbor sandboxes. Uses the pinned Slime fork for the agent loop, Modal sandboxes, and binary test-based rewards. Metrics also go to Trackio. Uses 2 `B300:1`, disaggregated.

Replaces the older agent_sandbox and multiturn tutorials.

Validation run: [Gym dashboard](https://modal-labs-helena-dev--training-gym-dashboard-fastapi-app.modal.run/training/proper-stage-21053b76c330?tab=summary), [Trackio dashboard](https://modal-labs-helena-dev--training-gym-trackio-dashboard.modal.run/?project=coding-agent&runs=weighted-aside-14be5a90da59&selected_run_id=984240c1e49c4668974ec7378f3162d0&selected_run=proper-stage-21053b76c330)

Built off of https://github.com/modal-projects/training-gym/pull/520 (add option for tutorials to be folders), with `main.py` and `dataset.py`

Removes `scripts/partition_swe_dataset.py`:

- `tutorials/coding_agent/dataset.py` handles SWE-rebench conversion, train/eval preparation, and mixed-reward selection as a preprocessing script.
- Reusable grouped splitting and balanced sampling in `modal_training_gym/common/dataset_partitioning.py` and operate directly on rows.
- `AgentTaskDataset` in `tutorials/coding_agent/main.py` reads the prepared JSONL for training. `DatasetConfig` is unchanged.

Screenshot 2026-09-08 at 1 48 50 PM

PR #506, opened 2026-09-04
Write torch_dist checkpoint shards from threads, and reject partial saves

zhouhelena1:
## Problem

Run `green-poodle-f5c489546992` (Qwen3.6-27B agentic smoke, one colocated 8×H200 node) trained both rollouts and then died at the final checkpoint save, twice in a row, with:

```
File ".../megatron/core/dist_checkpointing/strategies/filesystem_async.py", line 308, in write_preloaded_data_multiproc
File "/usr/lib/python3.12/multiprocessing/popen_fork.py", line 66, in _launch
    self.pid = os.fork()
BlockingIOError: [Errno 11] Resource temporarily unavailable
```

Megatron's `FileSystemWriterAsync.write_preloaded_data_multiproc` forks one worker process per write bucket, on every rank, at every save. In a container already running Megatron ranks, SGLang engines, Ray, and rollout thread pools, the fork can fail. The identical config saved fine the day before (`funny-interface-d2a700808798`), so this is environmental and intermittent, which is the worst kind. Upstream Megatron replaced the forked workers with threads for the same reason in NVIDIA/Megatron-LM#3633 (Feb 2026); the pinned image predates that.

The crashed save also exposed a second gap: `iter_0000001/` was left with `.distcp` shards and `common.pt` but no `.metadata`, and the slime launcher's `_is_complete_torch_dist_checkpoint` accepted it. The retry set `--load` to the partial directory, Megatron ignored it, and training silently restarted from `ref_load` at rollout 0.

## Changes

- **`common/megatron_patches/patch_ckpt_writer_threads.py`**: image-build patch that rewrites `write_preloaded_data_multiproc` to run the per-bucket workers in threads. Method name, signature, and the worker's queue protocol are unchanged, so `get_save_function_and_args` and the async caller are untouched. It is a no-op on a Megatron that already has upstream's threaded writer, idempotent, and refuses (fails the build) if the method has an unrecognized shape rather than guessing.
- Wired into the slime base image (Megatron-side patch list, survives the git overlay) and the miles image (guarded on the file existing, like the checkpoint-save patch).
- **Slime `_is_complete_torch_dist_checkpoint`** now requires `.metadata`, matching what the miles launcher already does.

## Verification

- Unit tests: the patch applied to an excerpt of pre-#3633 Megatron (`tests/testdata/megatron/filesystem_async.py.input`) removes the fork, keeps the signature, is idempotent, no-ops on the upstream threaded version, and refuses an unknown shape. The patched method is also executed against the unchanged worker with stub I/O: three buckets write and report through the queues, and a failing bucket propagates its exception. Completeness tests cover finished, partial, and missing directories plus the resume scan.
- The patch applies cleanly to the real pinned image at build time (`Patched filesystem_async.py to write checkpoint shards from threads`, right after the existing #3845 backport).
- Full suite: 929 passed, 1 skipped. ruff clean.
- End-to-end: a rerun of the same smoke with this fix is in progress; I'll post the outcome here.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/506" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/megatron_patches/patch_ckpt_writer_threads.py
+++ b/modal_training_gym/common/megatron_patches/patch_ckpt_writer_threads.py
@@ -1,0 +1,181 @@
+"""Write torch_dist checkpoint shards from threads instead of forked processes.
+
+Megatron's ``FileSystemWriterAsync.write_preloaded_data_multiproc`` forks one
+worker process per write bucket, on every rank, at every save. Inside a Modal
+container that is already running Megatron ranks, SGLang engines, Ray, and
+rollout thread pools, that ``os.fork()`` can fail with
+``BlockingIOError: [Errno 11] Resource temporarily unavailable``, which kills
+the save and the run with it (seen on Qwen3.6-27B, 2026-09-04).
+
+Upstream Megatron replaced the forked workers with threads for exactly this
+reason (NVIDIA/Megatron-LM#3633, "Add single-process checkpoint save to avoid
+forked multiprocessing"). This patch back-ports that behavior to the pinned
+image: the method keeps its name and signature, so ``get_save_function_and_args``
+and the async caller are untouched, and the per-bucket worker
+``write_preloaded_data`` is unchanged -- it already talks to its queues through
+``put``/``get``/``task_done``, which ``queue.Queue`` provides just as
+``multiprocessing`` queues do.
+
+Executed at image-build time via ``python3 <this file> [path]``.
+"""
+
+import pathlib
+import re
+import sys
+
+DEFAULT_PATH = (
+    "/root/Megatron-LM/megatron/core/dist_checkpointing/strategies/filesystem_async.py"
+)
+MARKER = "write(sync,threads)"
+
+METHOD_PATTERN = re.compile(
+    r"^    @staticmethod\n"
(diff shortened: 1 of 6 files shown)


zhouhelena1:
End-to-end result: the same smoke that died twice at the final save without this patch (`green-poodle-f5c489546992`) completed on the first attempt with it (`dynamic-raid-941a9eaf1379`, one colocated 8×H200 node, Qwen3.6-27B agentic recipe, 2 rollouts + 3 evals).

- Megatron logged `successfully saved checkpoint from iteration 1`; save took 275 s with the threaded writer versus 285 s with the forked writer on the last run that survived it (`funny-interface`, 2026-09-02), so no slowdown.
- `iter_0000001/` has 16 `.distcp` shards, `common.pt`, `.metadata`, `metadata.json`, and the tracker file reads `1`, i.e. a complete checkpoint by the new predicate.
- Zero tracebacks in the run log; Trackio upload messages from every process; app stopped cleanly.

The pinned image's Megatron is core 0.16.0rc0 at a 2026-02-14 commit, two weeks before upstream's #3633 landed, which is why it still forks.

zhouhelena1:
Folded into #505 at the author's request; closing.

PR #507, opened 2026-09-04
Render flattened multi-turn transcripts on the dashboard instead of a blank response

zhouhelena1:
## Problem

Opening a rollout of an agentic run (`dynamic-raid-941a9eaf1379`, the Qwen3.6-27B Harbor smoke) on the dashboard shows the prompt but no response. The data is recorded: every sample on the metadata volume has a 900 to 3000 character response. It is the display path that blanks it.

The agentic Slime fork appends later turns to `sample.response` verbatim, chat-template tokens included:

```
{assistant}<|im_end|>
<|im_start|>user
<tool_response>...</tool_response><|im_end|>
<|im_start|>assistant
<think>...</think>
<tool_call>...</tool_call><|im_end|>
```

At record time `parsed_response` is computed with the model's parser, which is single-turn: it keeps only the text after the last `<|im_start|>assistant`, and once the tool call is stripped that is `""`. At serve time `_apply_parsed` does `row["response"] = parsed["content"] or ""`, so the API returns an empty response, and `ConversationView` (no `trajectory_messages`, empty `response`) renders nothing. Same for every earlier run of this recipe and for `multiturn_slime`, whose dashboard changes were Trackio links only.

## Changes

- `training_rollout._transcript_messages` splits a flattened transcript into role-tagged messages; a `user` turn that is only a `<tool_response>` block becomes `tool` and is unwrapped. `_apply_parsed` stores them as `metadata.trajectory_messages` (a hook-supplied trajectory wins), never replaces a real response with empty parsed content, and skips the top-level `thinking` for transcripts since each turn carries its own. The frontend already renders `trajectory_messages` turn by turn with per-turn thinking toggles and tool calls, so no frontend change.
- `sample_extraction._sample_to_dict`: an oversized metadata tag that is a dict keeps its small entries instead of vanishing. The fork's `agentic` tag (turns, timings, `is_solved`, outcome) was being dropped because it also carries the episode's full prompt.
- `DASHBOARD_VERSION` 1 to 2 so `ensure_dashboard_deployed()` redeploys.

## Verification

- Applied `_apply_parsed` to the actual recorded rollout of `dynamic-raid-941a9eaf1379`: both samples split into `assistant` / `tool` / `assistant` turns with the tool output unwrapped, response kept, no stray top-level thinking.
- New tests cover the transcript split, the blank-response regression, a hook-supplied trajectory taking precedence, the unchanged single-turn path, and the metadata compaction. Existing route and CLI display tests pass unchanged.
- Full suite: 923 passed, 1 skipped. ruff clean.

Fixes the display for already recorded runs once the dashboard is redeployed (`training-gym setup`, or automatically on the next launch from a checkout with this change).

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/507" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/dashboard.py
+++ b/modal_training_gym/common/dashboard.py
@@ -17,7 +17,7 @@ DASHBOARD_PREVIEW_ENV_KEY = "TRAINING_GYM_DASHBOARD_PREVIEW"
 DASHBOARD_VERSION_ENV_KEY = "DASHBOARD_VERSION"
 
 # Bump when the deployed dashboard frontend or backend changes.
-DASHBOARD_VERSION = 1
+DASHBOARD_VERSION = 2
 
 
 def current_dashboard_version() -> str:
--- a/modal_training_gym/common/sample_extraction.py
+++ b/modal_training_gym/common/sample_extraction.py
@@ -667,6 +667,17 @@ def _sample_to_dict(
                 continue
             if _is_small_json_value(value):
                 metadata[key] = value
+            elif isinstance(value, dict):
+                # One oversized entry (an agent loop stashing its full prompt
+                # next to its turn counts and timings) should not erase the
+                # whole tag; keep the sub-entries that fit on their own.
+                compact = {
+                    sub_key: sub_value
+                    for sub_key, sub_value in value.items()
+                    if sub_value is not None and _is_small_json_value(sub_value)
+                }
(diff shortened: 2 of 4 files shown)


zhouhelena1:
Folded into #505 at the author's request; closing.

PR #510, opened 2026-09-07
Honor explicit single-rank checkpoint conversion overrides

nanjiangwill:
An explicit conversion TP1/PP1 was treated like an unspecified layout, so the converter still used every GPU on the node. Preserve explicit overrides, including `mtp_num_layers=0`, while keeping automatic conversion for unspecified layouts.

Validation: 25 conversion/checkpoint tests pass, covering implicit layout, explicit TP1 or PP1, sharded training with single-rank conversion, zero MTP, and incompatible expert parallelism. Ruff and diff checks pass. A Qwen4B conversion ran successfully with one process on Modal.

The example-specific template checklist does not apply to this SDK fix.

---

<a href="https://modal.devinenterprise.com/review/modal-projects/training-gym/pull/510" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/modal_training_gym/common/launcher_utils.py
+++ b/modal_training_gym/common/launcher_utils.py
@@ -188,7 +188,8 @@ def get_checkpoint_conversion_policy(
     attribute is populated (the model script already sources them).
 
     TP/PP come from the ``conversion_*`` overrides when set, else from the training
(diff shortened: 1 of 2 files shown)


joyliu-q (approved):
Thank you!

PR #517, opened 2026-09-08
Add SFT workload support to Slime recipes

omkaark:
Adds text SFT through the existing Slime model recipes: set `workload_type="sft"` and use `SFTDataset` with conversations or prompt/answer columns. RL remains the default. SFT reserves only actor GPUs and disables inference, critic, KL, and reward-based training work.

Includes data validation, workload identity in run summaries, tests, a one-GPU tutorial, and the three-PR release plan. `colocate=False` selects Slime's train-only execution path; it does not reserve an inference node.

Stack: **#517 → #518 → #519**. Review this first against `main`; #518 adds reliable SFT observability and #519 adds UI/UX. Each PR is based on its immediate parent. Rebased onto main `13f9e3ef` before implementation. Review model/dataset semantics before generated documentation.

Validation: focused SFT, recipe identity, dashboard schema, and CLI tests pass; lint/compile and documentation generation pass. The combined stack passes 980 Python tests (one skipped) and six frontend tests/build. A Qwen3-0.6B one-H100 proof and ten-step smoke completed with actor=1/inference=0/critic=0, saved checkpoints, and all ten optimizer metrics. Full integration evidence and remaining release checks are in #519 and `plan.md`.

Text SFT is the initial supported scope; multimodal/tool-call SFT and Miles SFT are excluded. Broad GPU model-validation CI was deliberately cancelled before execution to honor the one-H100 budget; snapshot CI awaits an authorized environment reviewer. Do not treat the cancelled matrix as a passing all-model validation.


docs-next/public/llms.txt
```diff
@@ -1,6 +1,8 @@
 # Training Gym
 
-Modal Training Gym is an open-source Python SDK for RL post-training on [Modal](https://modal.com).
+Modal Training Gym is an open-source Python SDK for distributed post-training on
```
omkaark: > for "RL and SFT" post-training

omkaark: keep similar to the original.

docs-next/public/llms.txt
```diff
@@ -49,6 +51,7 @@ Repo: https://github.com/modal-projects/training-gym
 - [GUI Grounding with Qwen3-VL-8B](https://gym.modal.dev/tutorials/computer_use)
 - [On-policy Distillation (OPD) Across Model Families](https://gym.modal.dev/tutorials/cross_tokenizer_distillation)
 - [Frontier-scale training](https://gym.modal.dev/tutorials/multinode)
+- [Supervised fine-tuning](https://gym.modal.dev/tutorials/sft)
```
omkaark: remove... this does not exist

docs-next/public/llms.txt
```diff
@@ -84,6 +87,7 @@ Repo: https://github.com/modal-projects/training-gym
 
 - [DatasetConfig](https://gym.modal.dev/reference/datasetconfig)
 - [HuggingFaceDataset](https://gym.modal.dev/reference/huggingfacedataset)
+- [SFTDataset](https://gym.modal.dev/reference/sftdataset)
```
omkaark: remove. this does not exist.

modal_training_gym/common/dataset.py
```diff
@@ -57,6 +59,7 @@ class DatasetConfig:
     # a separate DatasetConfig instance for offline eval (Toolathlon, BFCL) set
     # this False so resolvers don't invent a companion ``eval.*`` file.
     writes_eval_paths: bool = True
+    supports_sft: ClassVar[bool] = False
```
omkaark: I dont like this, remove, its redundant

modal_training_gym/common/dataset.py
```diff
@@ -247,6 +250,135 @@ def prepare(self, path: str, eval_paths: dict[str, str] | None = None) -> None:
...
+    Assistant turns are supervised targets. In-training evaluation is not
+    materialized because Slime's SFT rollout supports training data only.
+    """
+
+    messages_column: str = ""
+    input_key: str = "messages"
+    label_key: str = "label"
+    apply_chat_template: bool = False
+    writes_eval_paths: bool = False
+    supports_sft: ClassVar[bool] = True
```
omkaark: remove

docs-next/public/llms.txt
```diff
@@ -92,6 +93,7 @@ Repo: https://github.com/modal-projects/training-gym
 - [HuggingFaceDataset](https://gym.modal.dev/reference/huggingfacedataset)
 - [MultimodalDataset](https://gym.modal.dev/reference/multimodaldataset)
 - [OnlineRollout](https://gym.modal.dev/reference/onlinerollout)
+- [SFTDataset](https://gym.modal.dev/reference/sftdataset)
```
andrewhinh: i'm not super keen on having new dataset classes for every use case (e.g., MultiModalDataset). imo these should all live as params in HuggingDatasetConfig, but feel free to convince me otherwise

modal_training_gym/common/dataset.py
```diff
@@ -252,6 +252,145 @@ def write(self, path: str) -> None:
             ds.to_json(path, orient="records", lines=True)
 
 
+class SFTDataset(HuggingFaceDataset):
```
andrewhinh, line 255: wait are we only supporting SFT for HuggingFace datasets?

also, im not sure i understand what the difference between this and HuggingFaceDataset is.

docs-next/public/llms.txt
andrewhinh, line 96: hmm, i wonder if creating all these use-case specific dataset classes (i.e., this, OnlineRollout, MultimodalDataset) is good practice. as a user, i'd expect these to either be automatically inferred or exposed as flags.

btw this pertains to a latent issue rather than something this PR introduces.

modal_training_gym/frameworks/slime/launcher.py
```diff
@@ -387,6 +388,14 @@ def build_slime_app(
 
     SlimeRecipe._validate_custom_model_architecture(model)
     SlimeRecipe._validate_datasets(dataset, eval_dataset)
+
+    if slime.training_type == "sft" and dataset.apply_chat_template():
```
andrewhinh, line 392: to my point above, this seems like a sharp edge for a user to have to experience.

modal_training_gym/train_recipes/slime_recipe/recipe.py
```diff
@@ -167,6 +193,12 @@ class SlimeRecipe(BaseTrainRecipe):
             Megatron actor nodes.
         actor_num_gpus_per_node:
             GPUs per actor node.
+        num_gpus_per_node:
```
andrewhinh, line 196: hmm, im not a big fan of having separate parameters like this. even more confusingly, although this is for slime, there is already a [miles flag](https://miles.radixark.com/docs/diffusion/user-guide/cli-reference#cluster-topology) with this name. i wonder if we should instead make a `SFTRecipe`.

tutorials/sft.py
```diff
@@ -0,0 +1,37 @@
+# ---
```
andrewhinh, line 1: i think this is better suited as a new section in the Dataset guide.

andrewhinh (changes requested):
thanks for taking this on! it's badly needed.

plz address my and Devin's comments and lmk if you have any questions.

modal_training_gym/common/dataset.py
omkaark: this is a really good point

PR #518, opened 2026-09-08
Report SFT loss and targets without RL metrics

omkaark:
SFT now reports reduced optimizer loss and supervised conversation samples without producing RL reward, advantage, or generation statistics. This addresses the earlier successful training run whose dashboard stopped at 81/100 steps.

Metrics are batched into 100-step chunks using existing metadata storage. A checkpoint-side journal is written before background publication and reconciled after training. Samples reuse the existing rollout storage/export, capped at two conversations every ten steps plus the final step. Adds a loss-history read endpoint and specializes `run get --verbose`; no new metrics service or HTTP ingestion API.

Stack: **#517 → #518 → #519**. Base: `omkaark/sft`. Review only the diff against #517; #519 consumes this contract.

Validation: the combined stack passes 980 Python tests, one skipped. Includes a deterministic 100-step outage/recovery test, retry ordering, non-finite loss rejection, patch idempotence against pinned Slime fixtures, API coverage, and target preservation. Self-review/live testing also added regressions for stale status writes overwriting loss/progress and no-op RL timing/status noise; journal commits now tolerate in-flight checkpoint writers. Lint and compile checks pass.

Live one-H100 Qwen3-0.6B smoke: all ten scalar records, 10/10 actual progress, loss 2.2542 → 0.7857, both configured sample batches retained, and timings for all ten steps with no reward/inference/weight-sync phases. Full run/checkpoint evidence is in #519. Broad GPU model-validation CI was cancelled before execution to honor the budget; snapshot CI requires an authorized environment reviewer.

### DIFF
--- a/modal_training_gym/_dashboard.py
+++ b/modal_training_gym/_dashboard.py
@@ -84,6 +84,7 @@ from modal_training_gym.common.training_rollout import (
     TrainingRolloutSummary,
     _apply_parsed,
 )
+from modal_training_gym.common.training_steps import TrainingStep
 from modal_training_gym.utils.metadata import (
     bounded_gather_with_retries,
     vol_get as _metadata_vol_get,
@@ -186,6 +187,7 @@ PASSWORD_EXEMPT_PATHS = frozenset(
         DASHBOARD_VERSION_PATH,
         "/api/framework-status",
         "/api/training-rollouts",
+        "/api/training-steps",
         "/api/advantage-distributions",
         "/api/timing-events",
     }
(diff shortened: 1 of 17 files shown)


modal_training_gym/cli/run.py
```diff
@@ -602,33 +618,51 @@ def get_run(*, run_id: str, verbose: bool, json_output: bool) -> None:
                     params=None,
                 )
             )
-            if verbose
+            if verbose and summary.training_type != "sft"
+            else []
+        )
+        steps = (
```
andrewhinh, line 624: can you explain why the notion of steps in SFT differs from the RL case such that we need separate logic for it? it's not immediately obvious to me why only SFT would have the notion of steps.

modal_training_gym/common/reporting.py
```diff
@@ -120,7 +120,13 @@ def _step_progress(args: Any, rollout_id: int | None = None) -> dict[str, Any]:
     if rollout_id is None:
         current = 0
     else:
-        current = max(0, int(rollout_id) + 1)
+        if (
+            _arg_value(args, "loss_type") == "sft_loss"
+            or os.environ.get("TRAINING_GYM_TRAINING_TYPE") == "sft"
+        ):
+            current = max(0, int(rollout_id))  # since SFT counts starting 0
```
andrewhinh, line 127: but then shouldn't we just make RL count from 0 too?

modal_training_gym/common/reporting.py
```diff
@@ -182,11 +188,17 @@ def _enqueue(
     payload: dict[str, Any],
     *,
     timeout_seconds: float = _PHASE_TIMEOUT_SECONDS,
+    url: str | None = None,
 ) -> None:
     """Enqueue a framework-status payload with its request timeout."""
+    if os.environ.get("TRAINING_GYM_TRAINING_TYPE") == "sft" and payload.get(
+        "phase"
+    ) in {"weight_sync", "compute_log_probs", "offload_rollout", "evaluate_rollouts"}:
```
andrewhinh, line 196: wait we still get these from miles/slime even if we do SFT?

modal_training_gym/frameworks/slime/modal_helpers/patches/patch_sft_reporting.py
```diff
@@ -0,0 +1,60 @@
+import ast
```
andrewhinh, line 1: a reference to where this comes from (i.e., image pin, commit, file, and links to one or more of the three) would be greatly appreciated

modal_training_gym/_dashboard.py
andrewhinh, line 1: you'll likely have to bump the dashboard version, it's a const introduced in https://github.com/modal-projects/training-gym/pull/497

andrewhinh (changes requested):

PR #565, opened 2026-09-18
Mirror wandb/trackio metrics into a dashboard Metrics tab

joyliu-q:
Every scalar the framework logs through `wandb.log` now shows up in a **Metrics** tab on the run detail page — with any metric provider, no opt-out.

**Client** (`common/metric_mirror.py`): one `.pth` in the training image calls `bootstrap()`, and `TRAINING_GYM_METRIC_PROVIDER` alone picks the `wandb` surface:

- `dashboard` — new `DashboardMetricConfig`: W&B-shaped shim whose only sink is the dashboard (no account/key). **This is now the recipe default** (`SlimeRecipe.metrics` / `MilesRecipe.metrics`); `metrics=None` opts out of metric logging entirely.
- `trackio` — Trackio's existing adapter, which now also calls `mirror_log`.
- `wandb` — the real package, with `Run.log` teed via an import hook.

`MetricMirror` coalesces points per step (W&B `step`/`commit=False` semantics, nested dicts flattened to `a/b`, non-finite/non-scalar dropped) and ships them on the existing reporting queue every 2s, so they carry the per-run bearer token and drain at exit.

**Server** (`_dashboard.py`, `common/metric_series.py`):

```
POST /api/metric-points            {training_run_id, points: [{step, metrics}], final}
GET  /api/runs/{id}/metrics
     -> {series: {key: [[step, value], ...]}, stale}   # ≤1000 pts/key, min/max-preserving
```

One in-memory `{step: {key: value}}` table per run; dirty 1000-step chunks are written to `metric-series/{run}/chunk-NNNNNN-{writer}.json` (one file per dashboard replica, merged on read by per-key ingest time) by a 5s background loop, on `final`, on terminal status, and at shutdown. Last write wins per `(step, key)`.

**Frontend**: `RunMetricsPanel` — prefix-grouped `LineChart`s, key search, shared zoom, live polling while running.

**Docs**: metrics guide leads with the Metrics tab / `DashboardMetricConfig`.

## Checklist

Link to Devin session: https://modal.devinenterprise.com/sessions/57a8fff9ca5543d991140af69052521c
Open in Devin Desktop: https://modal.devinenterprise.com/desktop/session/57a8fff9ca5543d991140af69052521c?variant=devin
Requested by: @joyliu-q


docs-next/src/content/docs/guides/tools/metric.md
```diff
@@ -4,10 +4,33 @@ order: 2
...
+
+```python
+from modal_training_gym import DashboardMetricConfig, Qwen3_5_4B, Qwen3_5_4B_Recipe, TrainConfig
+
+config = TrainConfig(
+    model=Qwen3_5_4B(),
+    dataset=my_dataset,
+    recipe=Qwen3_5_4B_Recipe(
+        # ...
+        metrics=DashboardMetricConfig(project="my-rl-project"),
```
andrewhinh: hmm, i'd expect by default it would log to the dashboard without having to specify a value for `metrics`.

andrewhinh (commented):
i call out one behavior shift (by default, logging to the dashboard without specifying `metrics`: https://github.com/modal-projects/training-gym/pull/565#discussion_r4051302518)

o/w lgtm

docs-next/src/content/docs/guides/tools/metric.md
joyliu-q: Fair enough. Was thinking previously its a lot of data we may be storing and we do charge for storage, and users can opt into getting these metrics.

But maybe it should be an opt out flag instead.

PR #583, opened 2026-09-21
pin wandb on image

andrewhinh:
Wandb init still calls `wandb.util.generate_id` when `disable_random_suffix=False`, which was deprecated in `wandb 0.28.2`, so we pin to that.


joyliu-q (approved):
LGTM post conflict resolution

